PR #69, opened 2022-09-22
Improved seed selector

### LINKED ISSUE #41: Not able to set Seed above 2147483647 or below -2147483647
I have not been able to manually set the seed above 2147483647 or below -2147483647. The random seeds are usually outside this range.

NullSenseStudio:
Replaced seed selection with a random seed bool and manual seed string properties. The string property will allow seeds to be anything within a 32-bit uint (which appears to be the limit from pytorch's internals pytorch_lighting/utilities/seed.py seed_everything()) and fixes #41.

image

If the seed cannot be parsed to an int it will get a seed based on its hash. After an image is generated the seed property will get updated with the random seed or hash so someone may easily disable random seed to tweak settings when a desired seed is found.

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -101,6 +101,7 @@ class DreamTexture(bpy.types.Operator):
                     if area.type == 'IMAGE_EDITOR':
                         area.spaces.active.image = image
                 scene.dream_textures_progress = 0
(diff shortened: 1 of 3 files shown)


carson-katri:
Just tested and it works great!

PR #74, opened 2022-09-22
Fixed AttributeError with NoneType in context.area.regions

blastframe:
I was getting this error in the System Console. This is a fix.
```
 for region in context.area.regions:
AttributeError: 'NoneType' object has no attribute 'regions'
```

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -78,10 +78,12 @@ class DreamTexture(bpy.types.Operator):
         scene = context.scene
 
         def step_progress_update(self, context):
-            for region in context.area.regions:
-                if region.type == "UI":
-                    region.tag_redraw()
+            if hasattr(context.area,"regions"):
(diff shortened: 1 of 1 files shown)


operators/dream_texture.py
```diff
@@ -78,10 +78,12 @@ async def dream_texture(self, context):
         scene = context.scene
 
         def step_progress_update(self, context):
-            for region in context.area.regions:
-                if region.type == "UI":
-                    region.tag_redraw()
+            if hasattr(context.area,"regions"):
```
carson-katri, line 81: Small formatting nitpick, I’ll add an autoformat precommit hook at some point:
```suggestion
            if hasattr(context.area, "regions"):
```

carson-katri (changes requested):
Thanks for the fix. I think someone may have reported this, I’ll look for and close that issue if so.

Issue #42, opened 2022-09-20
Not releasing Vram until blender closes.


Multi21uk:
Windows 10 
3070ti  (516.94)
32gb ram
Blender 3.30

After running Dream Texture generation, the Vram is not released, disabling the ability to use cycles to view any preview/render due to lack of vram.

PR #81, opened 2022-09-23
Make Free Generator also empty cuda cache

NullSenseStudio:
I never noticed any VRAM releasing when activating Free Generator before making these changes.

`torch.cuda.empty_cache()` doesn't quite clear everything on its own and `generator.model.cpu()` seems to let it clear some more. Still had about 750MB of VRAM uncleared (according to `torch.cuda.memory_reserved()`) after that. I can't figure out how to clear the rest.

Unsure if `torch.cuda.empty_cache()` may error for a non-cuda system, someone else with appropriate hardware will have to test.

Not a perfect solution for #42, but it does seem to mostly do what it needs now.

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -203,6 +203,11 @@ class ReleaseGenerator(bpy.types.Operator):
 
     def execute(self, context):
         global generator
+        if generator:
+            generator.model.cpu()
         generator = None
+        gc.collect()
+        import torch
+        torch.cuda.empty_cache()
         context.scene.dream_textures_progress = 0
         return {'FINISHED'}


carson-katri:
Thanks for investigating this! I develop primarily on Mac, so it was tricky to figure this out because the memory is unified.

I’ll try to find a suitable solution for Apple Silicon to include in this before merging.

NullSenseStudio:
Also just thought it may be useful to run this automatically on `bpy.app.handlers.render_init` for people who don't realize the operator exists or just forgot to run it.

carson-katri:
This doesn't crash on Apple Silicon, but it has no effect. I was not able to find any similar function for MPS unfortunately.

NullSenseStudio:
Could move the generator to its own process and just kill it. Would also allow for easily cancelling in the middle of generating in case someone had their settings too high. Also may save blender from crashing in a few cases when the generator runs into issues.

PR #96, opened 2022-09-23
Generator subprocess

NullSenseStudio:
A replacement of #81 that keeps the generator within a subprocess so all platforms should be able to release its ram usage. This should fully fix #42 and seems to have also accidentally fixed #35 console spam.

Also allows for canceling in the middle of generating without the need to restart blender.

Still needs work to get progress and show steps so not quite ready to merge. I'd like you to give it a test before I bother with the rest.

PR #98, opened 2022-09-24
Remove dependency installation requirement

carson-katri:
Dependencies are now kept inside the addon at `.python_dependencies`, and can be distributed in the ZIP file. The addon will need to be distributed per-platform, something like:
* `dream_textures-darwin-arm64.zip`
* `dream_textures-win32-cuda.zip`
* ...

I will try to setup GitHub actions to package up the addon automatically for all platforms.

If you want to install dependencies in the main source code for local development, enable 'Developer Extras' in Blender preferences, then run 'Install Dependencies' in the preferences window.

For the changes in #96, the following will need to be added to the subprocess:
```python
site.addsitedir(absolute_path(".python_dependencies"))

import pkg_resources
pkg_resources._initialize_master_working_set()
```


carson-katri:
Just tested the latest artifact on Windows and it worked.

carson-katri:
@NullSenseStudio @blastframe
Would someone else be able to try it as well? You would want to remove the dependencies you already have installed in `C:\Program Files\Blender Foundation\...\site-packages` and then install the [addon artifact](https://github.com/carson-katri/dream-textures/actions/runs/3119724693). The actual ZIP file for the addon is nested one level deep, so unzip the file it downloads and install the ZIP it contained.

carson-katri:
I will need to create darwin-arm64 releases manually until GitHub actions adds Apple Silicon runners.

carson-katri:
If someone can verify that the artifact generated by GH Actions works on their machine I will get this merged.

NullSenseStudio:
I'll have it downloaded and tested in a few minutes.

NullSenseStudio:
Running perfectly fine. Also tested renaming the directory to make sure it wasn't somehow grabbing modules elsewhere, which just leads to `Process closed unexpectedly` thanks to it not finding them.

Only issue is github desktop thinking all these dependencies should be included in a commit. Just needs `.python_dependencies/` in `.gitignore` right?

blastframe:
Sorry I was late to seeing this. Great job @carson-katri and @NullSenseStudio !

carson-katri:
Unfortunately, the 2.5GB file exceeds the size limit for GitHub releases (2GB), so I either need to use a better compression algorithm or different file hosting service.

blastframe:
Is this any help?
https://git-lfs.github.com/

NullSenseStudio:
LZMA2 can drop it near 1.5GB
image

carson-katri:
@NullSenseStudio is blender able to install it directly or does it require `.zip`.

NullSenseStudio:
Looks like it does require `.zip`.
I'll see if it compresses differently for nesting.

carson-katri:
Ok, I think you can do a zip with no compression nested inside so it doesn’t effect the compression much.

NullSenseStudio:
Up to 2.16GB nested on normal compression level. Testing on ultra compression level now.

carson-katri:
I just created one that’s 1.6GB, a 7z of a zip with no compression.

NullSenseStudio:
Strange. I'm only getting down to 2GB on ultra 7z with non-compressed zip. Well as long as it works for you I won't worry about it.

carson-katri:
Here's what I ran:
```sh
zip -Z store -r dream_textures-win32-cuda.zip dream_textures -x "*.DS_Store" ".git/*" "readme_assets/*"
7z a -r dream_textures-win32-cuda dream_textures-win32-cuda.zip
```

carson-katri:
I need to add a `preload_models.py` operator to the setup before releasing, because the new version doesn’t work if you clear your `Users\user\.cache` folder. I should be able to release tomorrow.

NullSenseStudio:
Some of the troubleshooting tips are outdated and can be removed.

PR #134, opened 2022-09-25
Direct users who downloaded source to download release

NullSenseStudio:
Adds some error handling while loading dependencies within the subprocess and directs users who appear to have downloaded source to instead download from releases.
blender_ogSY8cKMSl

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -224,7 +224,7 @@ def main():
                 )
                 generator.load_model()
             except Exception as e:
-                writeException(True, str(e))
+                writeException(True, repr(e))
                 return
         writeInfo("Starting")
         
@@ -253,7 +253,7 @@ def main():
                         writeException(True, s) # consider all unknown exceptions to be fatal so the generator process is fully restarted next time
                         return
         except Exception as e:
-            writeException(True, str(e))
+            writeException(True, repr(e))
             return
         finally:
             sys.stderr = stderr
--- a/generator_process.py
+++ b/generator_process.py
@@ -113,30 +113,6 @@ class GeneratorProcess():
                 return
 
 def main():
-    from absolute_path import absolute_path
-    # Support Apple Silicon GPUs as much as possible.
-    os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
-    sys.path.append(absolute_path("stable_diffusion/"))
-    sys.path.append(absolute_path("stable_diffusion/src/clip"))
-    sys.path.append(absolute_path("stable_diffusion/src/k-diffusion"))
-    sys.path.append(absolute_path("stable_diffusion/src/taming-transformers"))
-
-    site.addsitedir(absolute_path(".python_dependencies"))
-    import pkg_resources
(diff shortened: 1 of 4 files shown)


generator_process.py
```diff
@@ -161,6 +137,38 @@ def writeException(fatal, e):
...
+
+        site.addsitedir(absolute_path(".python_dependencies"))
+        import pkg_resources
+        pkg_resources._initialize_master_working_set()
+
+        from stable_diffusion.ldm.generate import Generate
+        from omegaconf import OmegaConf
+        from PIL import ImageOps
+        from io import StringIO
+    except Exception as e:
```
carson-katri, line 157:
```suggestion
    except ModuleNotFoundError:
```
This would be more resilient than string checking if it works.

carson-katri (changes requested):

generator_process.py
NullSenseStudio: Yes it works. I'll still want it to catch any other exception just in case though.

generator_process.py
```diff
@@ -161,6 +137,41 @@ def writeException(fatal, e):
...
+
+        from stable_diffusion.ldm.generate import Generate
+        from omegaconf import OmegaConf
+        from PIL import ImageOps
+        from io import StringIO
+    except ModuleNotFoundError as e:
+        min_files = 10 # bump this up if more files get added to .python_dependencies in source
+                       # don't set too high so it can still pass info on individual missing modules
+        if not os.path.exists(".python_dependencies") or len(os.listdir()) < min_files:
+            e = "Python dependencies are missing. Click update to download full addon."
```
carson-katri, line 161: This needs an else branch so `writeException` works properly
```suggestion
            e = "Python dependencies are missing. Click update to download full addon."
        else:
            e = repr(e)
```

generator_process.py
```diff
@@ -161,6 +137,41 @@ def writeException(fatal, e):
...
+        writeException(True, e)
+        return
+    except Exception as e:
+        writeException(True, e)
```
carson-katri, line 165: I think this also needs to be converted to a str before being passed to `writeException`:
```suggestion
        writeException(True, repr(e))
```

operators/open_latest_version.py
```diff
@@ -19,6 +19,9 @@ def check_for_updates():
 
 def new_version_available():
     return not latest_version == VERSION
+def force_show_update():
```
carson-katri, line 22: I think I would prefer a separate global variable to keep track of a broken install. Then we can show a different button message and not change the `VERSION`, which could be used for something else in the future.

carson-katri (changes requested):

operators/open_latest_version.py
NullSenseStudio: Have that changed now.

NullSenseStudio: Glanced over the part about a different button message. I'll see what I can do about that next.

carson-katri (approved):
Looks great, thanks for the extra fixes and `repr(e)` changes.

PR #150, opened 2022-09-26
Add preload models step

carson-katri:
This is run every time the `GeneratorProcess` is started, so you don't need to do more than one thing in the preferences window. I also did not use the `preload_models.py` script from InvokeAI for more control over how it works.

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -191,6 +191,26 @@ def main():
             writeUInt(4,step)
         stdout.flush()
 
+    def preload_models():
+        import warnings
+        import transformers
+        transformers.logging.set_verbosity_error()
+
+        writeInfo("Preloading BERT tokenizer")
+        transformers.BertTokenizerFast.from_pretrained('bert-base-uncased')
+
+        writeInfo("Preloading `kornia` requirements")
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', category=DeprecationWarning)
+            import kornia
+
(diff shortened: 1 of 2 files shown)


carson-katri:
The progress is now shown for each model download:

Screen Shot 2022-09-26 at 7 09 08 PM

carson-katri:
@NullSenseStudio @blastframe 
There was a bug in this PR, but this one seems to be ok: https://github.com/carson-katri/dream-textures/actions/runs/3132180840
Can someone else give this a try before I release?

blastframe:
@carson-katri I'm happy to help if I can. Downloading now.

carson-katri:
@blastframe Thanks! if you can, delete the huggingface caches in `%HOMEPATH%\.cache` before generating your first image.

NullSenseStudio:
I can test too, but won't be until tomorrow if you're okay with waiting.

NullSenseStudio:
Worked perfectly for me. Also deleted torch within `.cache` to make sure the `checkpoint_liberty_with_aug.pth` file there would download too.

NullSenseStudio:
Is there any reason why the artifact is nested within a second .zip?

carson-katri:
The upload-artifact action creates a zip even if you pass it a zip to upload. And I need to compress it first or it compresses the directory bizarrely. I think it was trying to compress each file in the directory as its own artifact when I didn’t compress it first.

PR #186, opened 2022-09-29
Fix some install issues

NullSenseStudio:
Prevents a second installation of Dream Textures to be registered when one already is. Before it would cause some classes of the first installation to be unregistered, fail to fully register, and prevent the first installation from being able to be fully unregistered without a restart.

Also adds a warning when it appears to be installed from source zip in the preferences panel, previous "fix" #134 was insufficient as I forgot to account for the stable_diffusion directory not being included in source.
image


carson-katri:
Works great, thanks

Issue #192, opened 2022-09-29
Internal error, received image of wrong resolution 320x512, expected 512x512


lonedfx:
The new version(0.0.6) is giving errors when using the Init Image option with an image that isn't of the target dimension, regardless of whether the Fit to width/height option is selected:

For a 512x512 texture:
Internal error, received image of wrong resolution 320x512, expected 512x512 (with Fit to width/height)
Internal error, received image of wrong resolution 704x1024, expected 512x512 (without Fit to width/height)

I now have to select the size of the input image as the size of the texture by hand. That's not such a big deal obviously, but this was working fine in 0.0.5 without having to do so.

Silven88:
If your output is set to 512 x 512, and your source image resolution is 1024 x 2048, the error message will read "received image of wrong resolution 256x512, expected 512x512. 

This is telling you to set your output resolution to 256x512 in order to maintain the same aspect ratio as your source image. It works as long as you adjust your output resolution accordingly.

lonedfx:
@Silven88 thank you for your input, but no this is an actual regression, see my original message which clearly indicates that what worked in 0.0.5 no longer works in 0.0.6. In fact, the issue has already been fixed as part of revision #204, which is part of 0.0.7 milestone:

https://github.com/carson-katri/dream-textures/pull/204

"Fixes regression https://github.com/carson-katri/dream-textures/issues/192. The error check that is causing that regression has been swapped for a similar one that doesn't have expectations based on user input width and height."

carson-katri:
Closing as fixed in #204

PR #204, opened 2022-09-30
Subprocess IPC improvements

NullSenseStudio:
Redesigned backend to frontend IPC messages to be more easily modified and less prone to cause complicated issues like exceptions in the middle of writing an action to stdout. Added docstrings so other maintainers can better understand what a few parts do. Exception messages can now include a simplified user seen message and a stderr printed stacktrace, and has simplified integration when send_exception() is called within an except block. Message sending is also simplified by the send_action() function automatically packaging kwargs into json instead of individually writing things to stdout.

Fixes regression #192. The error check that is causing that regression has been swapped for a similar one that doesn't have expectations based on user input width and height.

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -11,8 +11,8 @@ from enum import IntEnum as Lawsuit, auto
 
 MISSING_DEPENDENCIES_ERROR = "Python dependencies are missing. Click Download Latest Release to fix."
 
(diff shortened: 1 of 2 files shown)


carson-katri (approved):
Nice refinements, thanks!

PR #229, opened 2022-10-02
Add "Dream Textures" render pass to Cycles

carson-katri:
This monkey patches a new render pass into cycles that can run stable diffusion each time a frame is rendered. It is especially useful for animations where you want to perform style transfer with stable diffusion.


carson-katri:
Memory consumption seems to be ok now. I'm not using 50+GB of swap after 50 frames anymore 😅

carson-katri:
Running cycles and dream textures can use a lot of VRAM, so if needed cycles can be run on CPU instead. When running through dream textures a lot of samples typically aren’t needed anyways.

render_pass.py
```diff
@@ -0,0 +1,111 @@
...
+                            def image_callback(event, shared_memory_name, seed, width, height, upscaled=False):
+                                self.update_stats("Dream Textures", "Pushing to render pass")
+                                # Only use the non-upscaled texture, as upscaling is currently unsupported by the addon.
+                                if not upscaled:
+                                    print(width, height)
+                                    print(len(render_pass.rect))
+                                    shared_memory = SharedMemory(shared_memory_name)
+                                    pixels = np.frombuffer(shared_memory.buf, dtype=np.float32)
+                                    reshaped = pixels.reshape((width * height, 4))
+                                    render_pass.rect.foreach_set(reshaped)
```
NullSenseStudio, line 56: This can error and freeze Blender with `TypeError: expected sequence size 133120, got 131072` when the output properties resolution isn't 512 or less in X or Y and when X or Y isn't a multiple of 64. Suggested changes in #245

ui/panels/render_properties.py
```diff
@@ -0,0 +1,49 @@
...
+
+        layout.prop(context.scene.dream_textures_render_properties_prompt, "strength")
+
+def render_properties_panels():
+    yield RenderPropertiesPanel
+    def get_prompt(context):
+        return context.scene.dream_textures_render_properties_prompt
+    space_type = RenderPropertiesPanel.bl_space_type
+    region_type = RenderPropertiesPanel.bl_region_type
+    panels = [
```
NullSenseStudio, line 34: May want to add a size panel so renders can be done in higher resolutions.

NullSenseStudio (changes requested):
Only a couple of issues.

ui/panels/render_properties.py
NullSenseStudio: Or potentially use the output properties resolution instead.

carson-katri: Using the output properties resolution would also resolve the issue from #245, correct? Although it would still need to require a multiple of 64, so perhaps that could simply show an error right away if detected, asking the user to update their output properties.

PR #233, opened 2022-10-03
Add Real-ESRGAN integration

carson-katri:
A new "AI Upscaling" panel has been added. I also refactored the generator process a bit to allow for multiple "Intents" (the two current ones being `PROMPT_TO_IMAGE` and `UPSCALE`)
Screen Shot 2022-10-02 at 9 48 38 PM

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -416,19 +416,17 @@ def main():
                 from realesrgan.archs.srvgg_arch import SRVGGNetCompact
                 # image = Image.open(args['input'])
                 image = cv2.imread(args['input'], cv2.IMREAD_UNCHANGED)
-                model = SRVGGNetCompact(num_in_ch=3, num_out_ch=3, num_feat=64, num_conv=32, upscale=4, act_type='prelu')
+                real_esrgan_model = SRVGGNetCompact(num_in_ch=3, num_out_ch=3, num_feat=64, num_conv=32, upscale=4, act_type='prelu')
                 netscale = 4
                 send_info("Loading Upsampler")
                 upsampler = RealESRGANer(
                     scale=netscale,
                     model_path=REAL_ESRGAN_WEIGHTS_PATH,
-                    dni_weight=1,
-                    model=model,
+                    model=real_esrgan_model,
                     tile=0,
                     tile_pad=10,
                     pre_pad=0,
-                    half=False,
-                    gpu_id=None
+                    half=not args['full_precision']
                 )
                 send_info("Enhancing Input")
                 output, _ = upsampler.enhance(image, outscale=args['outscale'])
--- a/generator_process.py
+++ b/generator_process.py
@@ -9,6 +9,7 @@ import traceback
 import numpy as np
 from enum import IntEnum
 import tempfile
+from multiprocessing.shared_memory import SharedMemory
 
 MISSING_DEPENDENCIES_ERROR = "Python dependencies are missing. Click Download Latest Release to fix."
 
@@ -22,7 +23,6 @@ class Action(IntEnum):
     STEP_IMAGE = 3
     STEP_NO_SHOW = 4
     EXCEPTION = 5
-    IMAGE_PATH = 6 # An image result that returns a path to a saved image instead of bytes
 
     @classmethod
     def _missing_(cls, value):
@@ -114,7 +114,7 @@ class GeneratorProcess():
         queue = self.queue
         callbacks = {
             Action.INFO: info_callback,
-            Action.IMAGE_PATH: image_callback,
+            Action.IMAGE: image_callback,
             Action.EXCEPTION: exception_callback
         }
 
@@ -124,7 +124,7 @@ class GeneratorProcess():
             tup = queue.pop()
             action = tup[0]
             callbacks[action](**tup[1])
-            if action == Action.IMAGE_PATH:
+            if action == Action.IMAGE:
                 break
             elif action == Action.EXCEPTION:
                 return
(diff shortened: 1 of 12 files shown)


NullSenseStudio:
Due to the stable diffusion branch change I've been getting `ImportError: DLL load failed while importing cv2: The specified module could not be found.` on line `from stable_diffusion.ldm.generate import Generate` while debugging, and oddly no exception is caught when starting Blender normally and just get `Process closed unexpectedly`.

For some reason Blender puts its `python3.dll` in Blender's root directory rather than in `blender/3.3/python/bin`. I've had no luck with changing the subprocess's cwd or modifying the PATH environment variable to help the subprocess find it.

So far only workarounds I've got are copying/symlink `python3.dll` into `blender/3.3/python/bin`, running the subprocess from a standard python runtime, or with some added code before that import:
```python
if sys.platform == 'win32':
    from ctypes import WinDLL
    WinDLL(r"C:\Blender\stable\blender-3.3.0+lts.0759f671ce1f\python3.dll")
```
We should be able to get a proper path for other systems by modifying bpy.app.binary_path and sending to the subprocess as a command line argument.

NullSenseStudio:
I should have time to do that later today. I'll also double check if the solution works for the msi installer and on steam.

NullSenseStudio:
PR has been made and can confirm that zip, msi, and steam versions all had the same issue and worked with the fix.

carson-katri:
@NullSenseStudio Can you give this a try before I merge it?

operators/dream_texture.py
```diff
@@ -124,15 +118,17 @@ def image_writer(seed, width, height, pixels, upscaled=False):
...
             scene.dream_textures_progress = step + 1
-            if pixels is None:
+            if shared_memory_name is None:
                 return # show steps disabled
             global last_data_block
             for area in screen.areas:
                 if area.type == 'IMAGE_EDITOR':
-                    step_image = bpy_image(f'Step {step + 1}/{scene.dream_textures_prompt.steps}', width, height, pixels)
+                    shared_memory = SharedMemory(shared_memory_name)
+                    shared_memory.close()
```
NullSenseStudio, line 130: shared_memory is closed before it can be used

generator_process.py
```diff
@@ -281,57 +348,101 @@ def update(self, n=1):
...
+                from realesrgan.archs.srvgg_arch import SRVGGNetCompact
+                # image = Image.open(args['input'])
+                image = cv2.imread(args['input'], cv2.IMREAD_UNCHANGED)
+                model = SRVGGNetCompact(num_in_ch=3, num_out_ch=3, num_feat=64, num_conv=32, upscale=4, act_type='prelu')
+                netscale = 4
+                send_info("Loading Upsampler")
+                upsampler = RealESRGANer(
+                    scale=netscale,
+                    model_path=REAL_ESRGAN_WEIGHTS_PATH,
+                    dni_weight=1,
```
NullSenseStudio, line 425: TypeError: RealESRGANer.__init__() got an unexpected keyword argument 'dni_weight'

generator_process.py
```diff
@@ -281,57 +348,101 @@ def update(self, n=1):
...
                     model=model,
-                    # These args are deprecated, but we need them to specify an absolute path to the weights.
-                    weights=weights,
-                    config=config,
-                    full_precision=args['full_precision']
+                    tile=0,
+                    tile_pad=10,
+                    pre_pad=0,
+                    half=False,
+                    gpu_id=None
```
NullSenseStudio, line 431: TypeError: RealESRGANer.__init__() got an unexpected keyword argument 'gpu_id'

generator_process.py
```diff
@@ -281,57 +348,101 @@ def update(self, n=1):
...
             try:
-                generator = Generate(
-                    conf=models_config,
+                from absolute_path import REAL_ESRGAN_WEIGHTS_PATH
+                import cv2
+                from realesrgan import RealESRGANer
+                from realesrgan.archs.srvgg_arch import SRVGGNetCompact
+                # image = Image.open(args['input'])
+                image = cv2.imread(args['input'], cv2.IMREAD_UNCHANGED)
+                model = SRVGGNetCompact(num_in_ch=3, num_out_ch=3, num_feat=64, num_conv=32, upscale=4, act_type='prelu')
```
NullSenseStudio, line 419: `model` variable name interferes with creating stable diffusion generator after upscaling

operators/dream_texture.py
```diff
@@ -87,20 +83,16 @@ def step_progress_update(self, context):
         bpy.types.Scene.dream_textures_progress = bpy.props.IntProperty(name="Progress", default=0, min=0, max=context.scene.dream_textures_prompt.steps + 1, update=step_progress_update)
         bpy.types.Scene.dream_textures_info = bpy.props.StringProperty(name="Info", update=step_progress_update)
```
NullSenseStudio, line 84: These properties are also needed in upscale.py. When the subprocess gets initiated from upscale the ui doesn't automatically update when these are changed until this operator runs.

NullSenseStudio (changes requested):
Here's some issues I ran into.

generator_process.py
carson-katri: Ah, I installed 0.3.0 locally, but set the windows requirements to 0.2.5.0.

carson-katri:
FYI, I added a "Full Precision" option to the upscaler but default to enabled. The precision options for the Dream Texture panel has been overhauled as well, and use an "Automatic" option now which can autodetect 16 series cards automatically.

NullSenseStudio:
Nice addition. Looks like everything is working well now.

Did there used to be a strength option for upscaling or am I misremembering?

NullSenseStudio (approved):
All previous issues seem fixed.

carson-katri:
Thanks for the reviews! I removed strength (for now) because it required two weight files to be downloaded which could be confusing, and it also is not available in 0.2.5.0 as far as I could tell (the `dni_weight` argument)?

PR #237, opened 2022-10-03
Fix cv2 ImportError

NullSenseStudio:
Due to the stable diffusion branch change cv2 now gets imported and doesn't play well with how Blender doesn't keep `python3.dll` in the same directory as `python.exe`. This fix allows the subprocess to be aware of its location.


carson-katri (approved):
Awesome, thanks!

PR #240, opened 2022-10-04
UI Refactor

carson-katri:
Moved each subsection into a subpanel, which matches Blender's UI better.
Screen Shot 2022-10-03 at 11 26 30 PM


carson-katri:
Model weight setup UI has been replaced with a open file dialog and automatic copying to the correct directory with `shutil`.

carson-katri:
I added a JSON-based prompt import/export, as well as some simple history management options.

Export from the history panel, then import in the Dream Texture panel.

carson-katri:
@NullSenseStudio Can I get your review on this before merging?

NullSenseStudio:
@carson-katri I should have some time to review soon. Thanks for the offer of being made co-maintainer but what all does that entail? This is the first repo I've done much help with and still learning things as I go.

carson-katri:
@NullSenseStudio Basically it would give you some extra permissions, such as merging pull requests, give approvals, close issues, etc. you could also make branches on this repo instead of your fork for PRs. If you don’t want that level of permissions yet that’s perfectly fine with me

NullSenseStudio:
@carson-katri  Alright I'll accept it. I'll just stick to managing branches in my own fork and let you handle merging for now until I understand things better. Reviewing these changes now.

NullSenseStudio (changes requested):
Good seeing all the clarifying information on the ui and simplified weights installation.

Only found two issues. First being that Blender thinks most of these properties are okay to animate, and I don't know what change is causing that.
image

carson-katri:
That can be disabled fairly easily 👍

NullSenseStudio (approved):
Works great

PR #245, opened 2022-10-05
add backend image resizing

NullSenseStudio:
Allows the backend to resize images before sending them to fix `TypeError: expected sequence size ...` when using the DT render pass and the output properties resolution doesn't match the images produced.

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -267,6 +267,9 @@ def main():
 
     def share_image_memory(image):
         nonlocal shared_memory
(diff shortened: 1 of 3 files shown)


carson-katri (changes requested):
I think it may be better to update the width/height args to match the output properties.

carson-katri:
Closing this as I implemented the width/height args to match the output properties in a95cf89843ead2d27866ae53441602b2dbf9c037

PR #261, opened 2022-10-06
Clear Modules from a Previous Install

NullSenseStudio:
Blender (at least on Windows) will only reload `__init__.py` when the user enables the addon and detects that the file has changed, leaving all other modules from the addon still loaded and leads to `__init__.py` attempting to use out-of-date modules after installing a new version.

This change will allow  `__init__.py` to clear those modules so they too get reloaded.

Installing a new version will still require that the previous version is disabled, removed from Blender, or when left enabled it is disabled and enabled again after installed.
blender_j6RmG2ZdAE
Version is misleading. bl_info gets updated when the new version is installed. It was still running 0.0.6 at the start of the gif

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -26,6 +26,14 @@ bl_info = {
 import bpy
 from bpy.props import IntProperty, PointerProperty, EnumProperty, BoolProperty
 import sys
+import os
+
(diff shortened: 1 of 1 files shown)


carson-katri (approved):
Nice QOL improvement, thanks

Issue #265, opened 2022-10-06
Setting "Precision" to a value other than "Automatic" causes the generator to load every time you generate


nikolatesla20:
**Is this a known issue?**
Not yet

 **Version** 
0.0.7

**Describe the bug**
If I choose (Half Precision) in the precision dropdown, then each time I click "generate" Dream Textures will say "Loading model" and it will take some time to load it. That slows down the next generation.

If I leave Precision on "Automatic" it does not appear to suffer from the problem

**To Reproduce**
1. Open Blender
2. Open Dream Textures
3. Set Precision to "Half Precision"
4. Generate an image. The first time through the model will have to load
5. After previous image is done, click "Generate" again. The model loads once again (slow)

**Expected behavior**
The model should not need to load each time you click generate.

**Screenshots**
![image](https://user-images.githubusercontent.com/3721469/194405169-e3105c89-58f0-424e-aace-131021fac10b.png)

**System (please complete the following information):**
- OS
  - [x] Windows
  - [ ] macOS
  - [ ] Other
- GPU
  - [x] NVIDIA
  - [ ] NVIDIA 16*XX*
  - [ ] Apple Silicon
  - [ ] AMD
- Addon Version
  - 0.0.7

**Additional context**

PR #274, opened 2022-10-07
More Subprocess Refactoring and Cancel Generate Operator

NullSenseStudio:
- Cancel Generate operator allows for stopping without killing the subprocess and reloading everything next time
- Added `send_intent()` method to `GeneratorProcess` that mirrors what `send_action()` does
- Subprocess state is now managed by the `Backend` class to clean up `main()`
- Subprocess now uses a background thread for reading stdin so new intents can be processed while one is already running
- Intents are split into their own generator functions that allow for lazy loading their dependencies and models, won't have to wait on stable diffusion models if you're only starting the subprocess for upscaling
- Includes a fix for #265

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -129,10 +129,10 @@ class HeadlessDreamTexture(bpy.types.Operator):
         try:
             next(generator_advance)
         except StopIteration:
-            remove_timer(context)
+            modal_stopped(context)
             return {'FINISHED'}
         except Exception as e:
-            remove_timer(context)
+            modal_stopped(context)
             raise e
         return {'RUNNING_MODAL'}
 
@@ -242,7 +242,7 @@ def modal_stopped(context):
         context.window_manager.event_timer_remove(timer)
         timer = None
         if not hasattr(context,'scene'):
-            context = bpy.context
+            context = bpy.context # modal context is sometimes missing scene?
         context.scene.dream_textures_progress = 0
         context.scene.dream_textures_info = ""
     global last_data_block
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -25,8 +25,7 @@ class DreamTexture(bpy.types.Operator):
 
     @classmethod
     def poll(cls, context):
-        global timer
-        return timer is None
+        return GeneratorProcess.can_use()
     
     def execute(self, context):
         history_entry = context.preferences.addons[StableDiffusionPreferences.bl_idname].preferences.history.add()
@@ -113,8 +112,7 @@ class HeadlessDreamTexture(bpy.types.Operator):
 
     @classmethod
     def poll(cls, context):
-        global timer
-        return timer is None
+        return GeneratorProcess.can_use()
 
     def invoke(self, context, event):
         weights_installed = os.path.exists(WEIGHTS_PATH)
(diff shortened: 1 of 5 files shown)


carson-katri:
> Subprocess now uses a background thread for reading stdin so new intents can be processed while one is already running
I assume this is meant for the stop generation action, but I also tried testing upscaling while the generator was running. This did not work:
```
  File "/Users/carsonkatri/Documents/Art/Add-ons/Custom/Blender/dream_textures/generator_process.py", line 518, in main
    back.main_loop()
  File "/Users/carsonkatri/Documents/Art/Add-ons/Custom/Blender/dream_textures/generator_process.py", line 480, in main_loop
    (intent, args) = self.queue.pop()

During handling of the above exception, another exception occurred:

  File "/Users/carsonkatri/Documents/Art/Add-ons/Custom/Blender/dream_textures/generator_process.py", line 255, in send_exception
    sys.exit(1)
  File "/Users/carsonkatri/Documents/Art/Add-ons/Custom/Blender/dream_textures/generator_process.py", line 522, in main
    back.send_exception()
  File "/Users/carsonkatri/Documents/Art/Add-ons/Custom/Blender/dream_textures/generator_process.py", line 525, in <module>
    main()
```
So perhaps we can either disable the UI for the upscaling panel while generator is running or look into a fix for this.

NullSenseStudio:
I wasn't quite able to replicate your error case but looks like I didn't make it properly wait if it had more than one intent in the queue. That should be okay now.

There is still an issue of multiple operators polling the action queue at the same time that will need addressed. I'll see what I can do about that later. Either disabling the operator while another is running or perhaps attaching a unique id to link each action to the intent that caused it.
```
  File "C:\Users\NullSense\AppData\Roaming\Blender Foundation\Blender\3.3\scripts\addons\dream-textures\generator_process.py", line 147, in upscale
    callbacks[action](**tup[1])
KeyError: 4
```

NullSenseStudio:
Blocking operators is the most practical for now. Intent session id tracking could be useful if the subprocess would split tasks up between multiple GPUs, but just an idea to save for later.

operators/dream_texture.py
```diff
@@ -233,25 +230,28 @@ def image_callback(shared_memory_name, seed, width, height, upscaled=False):
...
         context.window_manager.event_timer_remove(timer)
         timer = None
-
-def kill_generator(context=bpy.context):
-    GeneratorProcess.kill_shared()
-    remove_timer(context)
-    bpy.context.scene.dream_textures_progress = 0
-    bpy.context.scene.dream_textures_info = ""
+    if not hasattr(context,'scene'):
+        context = bpy.context # modal context is sometimes missing scene?
```
carson-katri, line 243: That's odd, since the scene is supposed to be in the global context, so all other contexts would be built on that (i assumed)

carson-katri (approved):
LGTM 👍

operators/dream_texture.py
NullSenseStudio: I may have mixed up where it was getting context from, or completely forgetting what caused the issue in the first place. There actually seems to be a problem with the context that `kill_generator()` gives it, the default context argument is somehow different from `bpy.context` when called. It should stop being an issue if I just leave the default as `None` and replace with `bpy.context` when that is the case.

Issue #205, opened 2022-09-30
Init Image [Attribute Error]


nyu777:
When I try to create an image from my own image with "init image", the process starts but I get an "Attribute Error".

windows10
Blender3.3LTS
v0.0.6
GPU Geforce3080
![SnapCrab_NoName_2022-9-30_16-53-22_No-00](https://user-images.githubusercontent.com/114099714/193222015-c7d6d9b4-242f-4920-b3fb-742e7a7d13ce.png)
![SnapCrab_NoName_2022-9-30_16-57-54_No-00](https://user-images.githubusercontent.com/114099714/193222038-d97c5502-a9f4-4ecd-9f55-a0d9b3171530.png)

ahoagmedia:
Same issue here, and also have an identical error when trying to use inpainting.  

Also on W10, Blender 3.3 LTS, v.0.0.6 with a 3080.

Smuzzies:
Just fixed it on mine, by removing a local copy of Python i was using for unrelated projects and then removing the folder - c:\user\name\appdata\python which contained an outdated version of PIL (pillow 9.0)

PR #317, opened 2022-10-16
Import Fixes

NullSenseStudio:
For some people the subprocess has managed to load external modules and cause odd issues. I've tracked down these ways that it may happen and should no longer be a problem with these changes.

1. PYTHONPATH environment variable: Becomes an issue when Blender is run with `--python-use-system-env` since the subprocess inherits Blender's environment variables. Now it'll always be removed from the subprocess's environment.
2. [User site](https://peps.python.org/pep-0370/): Blender will block use of the user site unless ran with `--python-use-system-env`, but the subprocess doesn't inherit this functionality. `-s` command line argument will prevent this.
3. Blender's site-packages: Some other addons may install their own packages there, and potentially packages installed from older versions of the addon may no longer be compatible. Reordered `sys.path` so modules will be looked for within the addon before any paths from the python interpreter.

Only reported issue that I know this will fix is #205 (loaded an outdated version of pillow, likely within user site like Smuzzies commented). Potentially others that haven't been properly tracked down too.

### DIFF
--- a/generator_process.py
+++ b/generator_process.py
@@ -59,7 +59,13 @@ _shared_instance = None
 class GeneratorProcess():
     def __init__(self):
         import bpy
-        self.process = subprocess.Popen([sys.executable,'generator_process.py',bpy.app.binary_path],cwd=os.path.dirname(os.path.realpath(__file__)),stdin=subprocess.PIPE,stdout=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


carson-katri (approved):
LGTM

PR #318, opened 2022-10-16
Upscale Seamless Fix

NullSenseStudio:
This adds an extra checkbox to upscaling that'll prevent artifacts around the edges of a seamless texture. 
blender_K34ckQV1kX


carson-katri (approved):
Works very nicely, thanks!

PR #323, opened 2022-10-17
render pass set pixels fix

NullSenseStudio:
Just found this bug moments after you merged. Most recent changes wasn't using the color transforms at all. Looks like inverting the color transform worked perfectly.
image


carson-katri (approved):
Ah, thanks for the catch!

PR #324, opened 2022-10-17
Advanced Presets

carson-katri:
This adds the ability to create and load presets for the 'Advanced' panel with the built-in Blender UI:
Screen Shot 2022-10-16 at 9 06 22 PM

Three defaults are included:
* Preview - lower steps for prototyping
  * Steps 25
  * Show Steps: Off
* Final - higher steps for final generations
  * Steps 50
  * Show Steps: Off
* Debug - shows each step
  * Steps 25
  * Show Steps: On

> **Note** Model weights were moved to `weights/stable-diffusion-v1.4`, so they are no longer in `stable_diffusion/models/ldm/stable-diffusion-v1`

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -44,6 +44,7 @@ from .operators.dream_texture import DreamTexture, kill_generator
 from .property_groups.dream_prompt import DreamPrompt
 from .operators.upscale import upscale_options
 from .preferences import StableDiffusionPreferences
+from .ui.presets import register_default_presets
 
 requirements_path_items = (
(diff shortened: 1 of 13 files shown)


NullSenseStudio:
I'll try to get a proper look at this tomorrow.

NullSenseStudio (approved):
Works great!

Might be useful to have an extra button somewhere to re-add the default presets if someone accidentally deleted theirs, or could just have each default preset individually checked and copied instead of checking for the parent directory in `register_default_presets()`.

NullSenseStudio (approved):
Perfect

PR #332, opened 2022-10-18
Add File Batch and Re-Enable Iterations

carson-katri:
This adds a new prompt preset called "File Batch", which runs the prompts from each line of a file:
Screen Shot 2022-10-18 at 2 27 00 PM

I also re-enabled the "Iterations" option, as it seems to be working fine now.

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -42,6 +42,7 @@ class DreamTexture(bpy.types.Operator):
         is_file_batch = context.scene.dream_textures_prompt.prompt_structure == file_batch_structure.id
         file_batch_lines = []
         if is_file_batch:
+            context.scene.dream_textures_prompt.iterations = 1
             file_batch_lines = [line for line in context.scene.dream_textures_prompt_file.lines if len(line.body.strip()) > 0]
             history_entries = [context.preferences.addons[StableDiffusionPreferences.bl_idname].preferences.history.add() for _ in file_batch_lines]
         else:
@@ -53,9 +54,6 @@ class DreamTexture(bpy.types.Operator):
             if is_file_batch:
                 history_entry.prompt_structure = custom_structure.id
                 history_entry.prompt_structure_token_subject = file_batch_lines[i].body
-        
-        if is_file_batch:
-            context.scene.dream_textures_prompt.iterations = 1
 
         def bpy_image(name, width, height, pixels):
             image = bpy.data.images.new(name, width=width, height=height)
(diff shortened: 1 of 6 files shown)


NullSenseStudio (changes requested):
Works great so far.

There is an issue with the history entries though. With iterations or batching it'll only store the seed from the last generated image, and with batching the prompt will just be an empty string. 

Do you think it would be better to save individual entries per each image or lists for prompt and seeds in a single entry?

carson-katri:
Entries are now created for each prompt, depending on batch/iterations. Also, iterations is forced to 1 when using file batch. This could potentially be lifted in the future.

operators/dream_texture.py
```diff
@@ -38,10 +38,24 @@ def invoke(self, context, event):
...
+        for i, history_entry in enumerate(history_entries):
+            for prop in context.scene.dream_textures_prompt.__annotations__.keys():
+                if hasattr(history_entry, prop):
+                    setattr(history_entry, prop, getattr(context.scene.dream_textures_prompt, prop))
+            if is_file_batch:
+                history_entry.prompt_structure = custom_structure.id
+                history_entry.prompt_structure_token_subject = file_batch_lines[i].body
+        
+        if is_file_batch:
+            context.scene.dream_textures_prompt.iterations = 1
```
NullSenseStudio, line 58: Iterations should be set to 1 before batch history entries are made, otherwise recalling one will include iterations that were not previously used.
```suggestion
        if is_file_batch:
            context.scene.dream_textures_prompt.iterations = 1
            file_batch_lines = [line for line in context.scene.dream_textures_prompt_file.lines if len(line.body.strip()) > 0]
            history_entries = [context.preferences.addons[StableDiffusionPreferences.bl_idname].preferences.history.add() for _ in file_batch_lines]
        else:
            history_entries = [context.preferences.addons[StableDiffusionPreferences.bl_idname].preferences.history.add() for _ in range(context.scene.dream_textures_prompt.iterations)]
        for i, history_entry in enumerate(history_entries):
            for prop in context.scene.dream_textures_prompt.__annotations__.keys():
                if hasattr(history_entry, prop):
                    setattr(history_entry, prop, getattr(context.scene.dream_textures_prompt, prop))
            if is_file_batch:
                history_entry.prompt_structure = custom_structure.id
                history_entry.prompt_structure_token_subject = file_batch_lines[i].body
```

NullSenseStudio (changes requested):
Just one more issue. The rest is working great.

PR #333, opened 2022-10-18
Fix InpaintAreaBrush behavior

carson-katri:
This replaces the manual stroke operator with a `GizmoGroup` that updates the "Draw" brush's blend mode when the "Mark Inpaint Area" brush is selected/de-selected.

Another hack, but this one seems to actually work well.

### DIFF
--- a/operators/inpaint_area_brush.py
+++ b/operators/inpaint_area_brush.py
@@ -1,62 +1,22 @@
 import bpy
-import math
 
-is_painting = False
-last_stroke = None
-time = 0
-paint_event = None
+reset_blend_mode = 'MIX'
+class InpaintAreaBrushActivated(bpy.types.GizmoGroup):
+    bl_idname = "dream_textures.inpaint_area_brush_activated"
+    bl_label = "Inpaint Area Brush Activated"
+    bl_space_type = 'IMAGE_EDITOR'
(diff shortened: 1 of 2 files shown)


operators/inpaint_area_brush.py
```diff
@@ -1,62 +1,32 @@
...
+        global reset_blend_mode
+        global reset_curve_preset
+        global reset_strength
+        reset_blend_mode = bpy.data.brushes["TexDraw"].blend
+        reset_curve_preset = bpy.data.brushes["TexDraw"].curve_preset
+        reset_strength = bpy.data.brushes["TexDraw"].strength
+        def set_blend():
+            bpy.data.brushes["TexDraw"].blend = "ERASE_ALPHA"
+            bpy.data.brushes["TexDraw"].curve_preset = "CONSTANT"
+            bpy.data.brushes["TexDraw"].strength = 1.0
```
NullSenseStudio, line 23: Inpaint can use the wrong tool if the draw tool wasn't selected
```suggestion
            bpy.data.brushes["TexDraw"].strength = 1.0
            bpy.ops.paint.brush_select(image_tool='DRAW', toggle=False)
```

NullSenseStudio (changes requested):
Works great! Even makes undo/redo work better. Only one small issue.

PR #334, opened 2022-10-19
Init Image Redesign and Outpainting

carson-katri:
* The init image panel was redesigned to better support multiple types of image manipulation:

Screen Shot 2022-10-19 at 1 26 18 PM

* 'Replace' option for inpainting allows the masked area to be replaced by noise
* 'Prompt' mask source uses CLIP segmentation to create an inpainting mask from a text prompt
* Outpainting support added, with a simple stretch/blur fill used to provide backing color information. The 'Replace' option could also be used to simply fill the new area with noise.

> **Note** A new git dependency was added for `clipseg` in the latest InvokeAI changes.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -77,7 +77,7 @@ def register():
         return ""
     bpy.types.Scene.dream_textures_history_selection = IntProperty(default=1)
     bpy.types.Scene.dream_textures_history_selection_preview = bpy.props.StringProperty(name="", default="", get=get_selection_preview, set=lambda _, __: None)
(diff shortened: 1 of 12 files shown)


NullSenseStudio:
Went to try out #335 but ran into a couple of issues detailed and fixed in #344

carson-katri:
Thanks, let me know if you notice anything that's broken or missing in this redesign.

NullSenseStudio (approved):
Looks like everything is working as expected and doesn't seem to be anything missing.

PR #344, opened 2022-10-22
clipseg fix

NullSenseStudio:
Had an issue that's already been addressed upstream: [ModuleNotFoundError: No module named 'models.clipseg'](https://github.com/invoke-ai/InvokeAI/issues/1150), and ran into another issue with a different module causing the subprocess to hang when attempting to generate.

### DIFF
--- a/generator_process/__init__.py
+++ b/generator_process/__init__.py
@@ -293,10 +293,14 @@ def main():
         import pkg_resources
         pkg_resources._initialize_master_working_set()
 
+        # Import specific modules that cause subprocess to hang if first imported after the background thread is started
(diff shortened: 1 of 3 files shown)


carson-katri (approved):
Works well on macOS too, thanks!

rexorp:
Can I get a step by step for adding the clipseg dependency models? I have downloaded the clipseg_weights.zip file but I don't know where to unzip it to. I have tried putting them in the weights or model folder(s) next to where the other models are/were but nothing seems to work for me. Looking for more help I just keep going in circles.

PR #348, opened 2022-10-24
DreamStudio integration

carson-katri:
This adds backend switching, with options "Local" and "DreamStudio" currently. In the future, other backends could be supported, such as an existing Stable Diffusion installation, HuggingFace's diffusers library, or something else.

DreamStudio setup is completely optional. However, releases will be made that only contain the dependencies necessary to run on DreamStudio (which will be much lighter weight). This will make the addon more accessible for lower-end hardware.


carson-katri:
This includes a very experimental implementation of realtime viewport rendering, but it is not currently accessible in the addon. Realtime viewport is not planned for 0.0.8.

requirements-dreamstudio.txt
```diff
@@ -0,0 +1 @@
+stability-sdk
```
NullSenseStudio, line 1: Either need to include `realesrgan` or disable upscaling panel.

generator_process/intents/prompt_to_image.py
```diff
@@ -196,4 +196,76 @@ def update(self, n=1):
...
+            # guidance_models: List[str] = None,
+        )
+
+        for answer in answers:
+            for artifact in answer.artifacts:
+                if artifact.finish_reason == interfaces.gooseai.generation.generation_pb2.FILTER:
+                    self.send_exception(False, "Your request activated DreamStudio's safety filter. Please modify your prompt and try again.")
+                if artifact.type == interfaces.gooseai.generation.generation_pb2.ARTIFACT_IMAGE:
+                    response = Image.open(io.BytesIO(artifact.binary))
+                    image_writer(response, seed)
```
NullSenseStudio, line 268: Iterations returns same seed for all images.
```suggestion
                    image_writer(response, artifact.seed)
```

NullSenseStudio (changes requested):
Source image will discard the image if `Color Correct` is checked. Inpainting and outpainting are generating no different from the modify option.

.github/workflows/package-release.yml
```diff
@@ -48,10 +48,64 @@ jobs:
...
+        uses: thedoctor0/zip-release@main
+        with:
+          type: zip
+          filename: dream_textures-macos-dreamstudio.zip
+          exclusions: '*.git*'
+      - name: Archive and upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: dream_textures-macos-intel-dreamstudio
+          path: dream_textures-macos-intel-dreamstudio.zip
```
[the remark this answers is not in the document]
carson-katri: Changed checks to look for contents of `stable_diffusion`, not its existence.

requirements-dreamstudio.txt
carson-katri: Disabled upscaling panel when the DreamStudio backend is selected, as well as other unsupported options.

operators/dream_texture.py
```diff
@@ -191,7 +195,12 @@ def step_progress_update(self, context):
         bpy.types.Scene.dream_textures_info = bpy.props.StringProperty(name="Info", update=step_progress_update)
         
         info("Waiting For Process")
-        generator = GeneratorProcess.shared()
+        if len(backend_options(self, context)) <= 1:
+            headless_prompt.backend = backend_options(self, context)[0]
```
NullSenseStudio, line 199: `TypeError: bpy_struct: item.attr = val: expected a string enum, not tuple`
```suggestion
            headless_prompt.backend = backend_options(self, context)[0][0]
```

generator_process/__init__.py
```diff
@@ -302,8 +317,9 @@ def main():
             # Couldn't track down exactly where it was hanging but it's good enough.
             from skimage import transform
```
NullSenseStudio, line 318: Forgot to previously test with only dream studio dependencies. Scikit-image is not a part of them.
```suggestion
            if os.path.exists(absolute_path(".python_dependencies/skimage")):
                from skimage import transform
```

generator_process/registrar.py
```diff
@@ -1,13 +1,58 @@
...
+                return True
+            case BackendTarget.STABILITY_SDK:
+                return False
+    
+    def upscaling(self):
+        match self:
+            case BackendTarget.LOCAL:
+                return True
+            case BackendTarget.STABILITY_SDK:
+                return False
```
NullSenseStudio, line 45: Seamless should also have its own backend check since DreamStudio does not support it.

NullSenseStudio (changes requested):
Changes looking good so far. Just a few more issues I've found.

carson-katri:
Thank you for the very thorough review!

PR #350, opened 2022-10-25
Model switching

carson-katri:
Adds a more advanced model importing UI to preferences, and adds a dropdown to select a model in the panel:
Screen Shot 2022-10-24 at 8 17 12 PM
Screen Shot 2022-10-24 at 8 16 59 PM
When switching, the `Generate` class will automatically be recreated.


NullSenseStudio (approved):
Works good. I've found no issues.

PR #369, opened 2022-10-30
tqdm decorator fix

### LINKED ISSUE #342: Ubuntu 22.04 TQDM object has no attribute "send_info"
I know Linux is not supported but I gave it a go. Something seems to be wrong with the progress bar module.

I had to manually install dependencies as I never got a button opencv, einops, torch_lightning by pointing to the version of Python Blender uses /usr/bin/python3.10 -m pip install MODULE.
Now when I enter a prompt it waits for a couple seconds thinking then spits the tqdm object has no attribute "send_info" error

NullSenseStudio:
Looks like I had accidentally shadowed a variable when reorganizing with the Backend class, which is causing a crash when attempting to download certain models before generating.

Fixes #342

### DIFF
--- a/generator_process/intents/prompt_to_image.py
+++ b/generator_process/intents/prompt_to_image.py
@@ -88,12 +88,12 @@ def prompt_to_image(self):
             self.send_info(f"Loading {model_name}")
 
         def update_decorator(original):
-            def update(self, n=1):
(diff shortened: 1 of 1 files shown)


carson-katri:
Thanks for the quick fix.

PR #370, opened 2022-10-31
Render Pass Fix

NullSenseStudio:
A regression was introduced when color correction was added, causing the render pass to use it when it shouldn't be, and other odd bugs cascading such as the frontend during OCIO inversion receiving that color corrected image while its backend crashes attempting to open an already closed shared_memory object.

These changes make sure that the render pass does not use the prompt_to_image color correction and keeps later stages from running until the previous one is entirely completed to prevent race conditions.


carson-katri (approved):
LGTM

PR #384, opened 2022-11-03
Linux Fixes

NullSenseStudio:
Main included fix is for `SSLCertVerificationError` while importing `kornia`. Made it so the base `ssl` module can use `certifi`'s `cacert.pem` file. The Blender 3.3.1 version I used has a broken `ssl` module that points to certs which do not exist, and I doubt this is the only version with the same problem.
```python
>>> import ssl
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(
	cafile=None,
	capath=None,
	openssl_cafile_env='SSL_CERT_FILE',
	openssl_cafile='/home/sybren/buildbot-builder/linux_glibc217_x86_64_cmake/build_deps/deps/Release/ssl/cert.pem',
	openssl_capath_env='SSL_CERT_DIR',
	openssl_capath='/home/sybren/buildbot-builder/linux_glibc217_x86_64_cmake/build_deps/deps/Release/ssl/certs'
)
>>> import os
>>> import certifi
>>> os.environ['SSL_CERT_FILE'] = certifi.where()
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(
	cafile='/home/nullsense/blender/blender-3.3.1/3.3/python/lib/python3.10/site-packages/certifi/cacert.pem',
	capath=None,
	openssl_cafile_env='SSL_CERT_FILE',
	openssl_cafile='/home/sybren/buildbot-builder/linux_glibc217_x86_64_cmake/build_deps/deps/Release/ssl/cert.pem',
	openssl_capath_env='SSL_CERT_DIR',
	openssl_capath='/home/sybren/buildbot-builder/linux_glibc217_x86_64_cmake/build_deps/deps/Release/ssl/certs'
)
```
After the change `urllib` won't fail on every https request and correctly downloads `checkpoint_liberty_with_aug.pth`. If there ends up being more issues with SSL verification it's likely fixable by updating the `certifi` module via pip.

Had an issue where for some reason pip was installed into the user site packages so ensurepip would not install it to Blender's `site-packages` directory. Made it so that should no longer accidentally happen.

Removed model preloading from the install dependencies operator since that's taken care of in `prompt_to_image.py` and just errors here.

Made it so the subprocess's stdout isn't discarded and is instead the same as stderr so it's still visible in the terminal. The model loading step will just print its error information so it was getting completely lost and leading to harder to debug issues.

This PR is not ready to merge yet since I'd like to spend more time trying to find more bugs to fix.

### DIFF
--- a/operators/install_dependencies.py
+++ b/operators/install_dependencies.py
@@ -40,7 +40,7 @@ def install_pip(method = PipInstall.STANDARD):
                 whl = package.wheel_path
             wheels[name] = whl
         pip_whl = os.path.join(wheels['pip'], 'pip')
-        subprocess.run([sys.executable, pip_whl, "install", *wheels.values(), "--upgrade", "--no-index", "--no-deps", "--no-cache-dir", "--target", absolute_path(".python_dependencies")])
+        subprocess.run([sys.executable, pip_whl, "install", *wheels.values(), "--upgrade", "--no-index", "--no-deps", "--no-cache-dir", "--target", absolute_path(".python_dependencies")], check=True)
         return
     
     # STANDARD or USER_SITE
--- a/operators/install_dependencies.py
+++ b/operators/install_dependencies.py
@@ -1,14 +1,21 @@
 import bpy
 import os
+import site
 import sys
 import sysconfig
 import subprocess
 import requests
 import tarfile
+from enum import IntEnum
 
 from ..absolute_path import absolute_path
 
-def install_pip(user_site=False):
+class PipInstall(IntEnum):
+    DEPENDENCIES = 1
+    STANDARD = 2
+    USER_SITE = 3
+
+def install_pip(method = PipInstall.STANDARD):
     """
     Installs pip if not already present. Please note that ensurepip.bootstrap() also calls pip, which adds the
     environment variable PIP_REQ_TRACKER. After ensurepip.bootstrap() finishes execution, the directory doesn't exist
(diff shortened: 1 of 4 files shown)


NullSenseStudio:
Forgot to mention I had also removed the `urlopen` decorator since it did not seem to help with the SSL issue I encountered, and I'd rather not have a potentially unsafe implementation like that.

I've now changed it so the `certifi` cert is always used on the backend rather than just during the model preloading phase in case if there are ever other methods using the standard SSL module.

The install dependencies operator will no longer attempt to preload models since that is handled on the backend and will always just error here.

Fixed a regression that made it so the missing dependencies warning wouldn't appear.

Snap installs programs under a read-only filesystem so now ensurepip can install to the user site when bootstrapping normally fails, and skips copying header files into the include directory.

operators/install_dependencies.py
```diff
@@ -22,33 +21,55 @@ def install_pip():
...
+        open(python_devel_tgz_path, 'wb').write(response.content)
+        python_devel_tgz = tarfile.open(python_devel_tgz_path)
+        def members(tf):
+            prefix = f"Python-{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}/Include/"
+            l = len(prefix)
+            for member in tf.getmembers():
+                if member.path.startswith(prefix):
+                    member.path = member.path[l:]
+                    yield member
+        python_devel_tgz.extractall(path=python_include_dir, members=members(python_devel_tgz))
```
NullSenseStudio, line 70: @carson-katri What exactly is this for? I've noticed no issues during installing and running without this step.

carson-katri: It fixed an error I had when initially developing, but I don’t remember exactly why. Perhaps needed when building from source on certain platforms.

carson-katri:
Do you know if there’s a bug filed for Blender about the OpenSSL paths?

NullSenseStudio:
I don't know. I'll try to look.

NullSenseStudio:
Can't find anything open or recent. I'll see about testing a beta build and file a report if it's still present.

NullSenseStudio:
Made it so Install Dependencies can now install and use pip in .python_dependencies, blender site-packages, and user site-packages (attempts in that order). I didn't like the user site-packages workaround since that takes precedence in sys.path and may interfere with a separate python interpreter running its pip module, but I'm keeping it in case if both alternatives fail.

NullSenseStudio:
I believe it's ready. I've tested on Windows to make sure it didn't accidentally cause issues there and it worked like normal.

carson-katri (approved):
LGTM 👍

PR #387, opened 2022-11-05
Add support for inpainting model and other model types import

### LINKED ISSUE #358: RunwayML Inpainting Model Support
**Is your feature request related to a problem? Please describe.**
RunwayML released a [model for inpainting](https://huggingface.co/runwayml/stable-diffusion-v1-5#training). InvokeAI has added support for this model in https://github.com/invoke-ai/InvokeAI/pull/1243. It should be usable when a "Source Image" is selected.

**Describe the solution you'd like**
Inpainting models will have a separate directory in the `weights` folder. They can then be selected when the backend is `LOCAL` and `use_init_img` is enabled.

**Describe alternatives you've considered**
The model could be selected like any other model in the `model` dropdown. The issue with this, is that the model requires different configuration, and detecting if a model is for inpainting could be problematic.

carson-katri:
Resolves #358
The model import UI was refactored to allow for various types of weights to be imported:
Screenshot 2022-11-05 at 1 58 46 PM
When importing, you can select which type of weights file is being imported in the sidebar.
Screenshot 2022-11-05 at 1 59 44 PM

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -43,6 +43,7 @@ from .property_groups.dream_prompt import DreamPrompt
 from .operators.upscale import upscale_options
 from .preferences import StableDiffusionPreferences
 from .ui.presets import register_default_presets
+from .absolute_path import REAL_ESRGAN_WEIGHTS_PATH
 
 requirements_path_items = (
     # Use the old version of requirements-win.txt to fix installation issues with Blender + PyTorch 1.12.1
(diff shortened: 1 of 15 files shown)


carson-katri:
You can find autoencoders to try here: https://huggingface.co/stabilityai/sd-vae-ft-mse

NullSenseStudio (changes requested):
Currently does not use the selected VAE. Prompt mask is broken due to its path pointing to a directory and there is no way to select the weights. Alpha channel mask is also broken, seems the backend isn't getting an image with transparency and the masked region is just replaced with black.

carson-katri:
`clipseg` dependency was updated to match InvokeAI changes, so it will need to be installed again.

NullSenseStudio (changes requested):
Previous issues are all fixed.

Looks like `requirements.txt` was renamed to `requirements-base.txt`.

Line 110 in `generator_process/intents/prompt_to_image.py` needs to be changed from
`from clipseg_models.clipseg import CLIPDensePredT` to `from clipseg.clipseg import CLIPDensePredT` due to that dependency change. Can't make a suggestion on that line as it's too far from any edits.

PR #396, opened 2022-11-16
Show the seed in history to easier match generated images to their promts

GottfriedHofmann:
When creating images, their are stored as a datablock with only the seed as name. Thus it would be very helpful to also have the seed in the history to make it easier going back to promt and fine-tuning.

grafik


carson-katri (approved):
Nice improvement, thanks!

PR #403, opened 2022-11-22
UI enhancement: When recalling a promt, open the corresponding image …

GottfriedHofmann:
…in the image editor when there is an image with the same name as the seed value in the promt.

When there is an image datablock with the same name as the seed from the promt to be recalled, open it in the image editor. When there is more than one datablock with the seed in it's name (like 12345.001, 12345.002 etc.) do not change the image open in the image editor and give a notification to the user.

https://user-images.githubusercontent.com/2581621/203351255-47cb1dba-d63b-4fe0-aeb1-a415fdd8af01.mp4

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -98,7 +98,7 @@ class DreamTexture(bpy.types.Operator):
                 scene.dream_textures_prompt.seed = str(seed) # update property in case seed was sourced randomly or from hash
                 # create a hash from the Blender image datablock to use as unique ID of said image and store it in the prompt history
                 # and as custom property of the image. Needs to be a string because the int from the hash function is too large
-                image_hash = str(hash(image))
+                image_hash = hashlib.sha256((np.array(image.pixels) * 255).tobytes()).hexdigest()
                 image['dream_textures_hash'] = image_hash
                 scene.dream_textures_prompt.hash = image_hash
                 history_entries[iteration].seed = str(seed)
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -1,6 +1,7 @@
 import sys
 import bpy
 import os
+import hashlib
 import numpy as np
 from multiprocessing.shared_memory import SharedMemory
 
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -96,8 +96,14 @@ class DreamTexture(bpy.types.Operator):
                     if area.type == 'IMAGE_EDITOR':
                         area.spaces.active.image = image
                 scene.dream_textures_prompt.seed = str(seed) # update property in case seed was sourced randomly or from hash
+                # create a hash from the Blender image datablock to use as unique ID of said image and store it in the prompt history
+                # and as custom property of the image. Needs to be a string because the int from the hash function is too large
+                image_hash = str(hash(image))
+                image['dream_textures_hash'] = image_hash
+                scene.dream_textures_prompt.hash = image_hash
                 history_entries[iteration].seed = str(seed)
                 history_entries[iteration].random_seed = False
+                history_entries[iteration].hash = image_hash
                 iteration += 1
         
         def view_step(step, width=None, height=None, shared_memory_name=None):
--- a/operators/view_history.py
+++ b/operators/view_history.py
@@ -41,20 +41,23 @@ class RecallHistoryEntry(bpy.types.Operator):
                 setattr(context.scene.dream_textures_prompt, prop, getattr(selection, prop))
             # when the seed of the promt is found in the available image datablocks, use that one in the open image editor
             # note: when there is more than one image with the seed in it's name, do nothing. Same when no image with that seed is available.
-            if prop == 'seed':
-                seed_string = str(getattr(selection, prop))
-                image_names_no_suffix = [i.name.split('.')[0] for i in bpy.data.images]
-                image_names_no_suffix_seed_count = image_names_no_suffix.count(seed_string)
-                if image_names_no_suffix_seed_count == 1:
-                    print("Image for seed " + seed_string + " is unique. Selecting image datablock.")
+            if prop == 'hash':
+                hash_string = str(getattr(selection, prop))
+                existing_image = None
+                # accessing custom properties for image datablocks in Blender is still a bit cumbersome
+                for i in bpy.data.images:
+                    try:
+                        # this will fail for images without the dream_textures_hash custom property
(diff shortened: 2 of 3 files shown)


operators/view_history.py
```diff
@@ -39,6 +39,22 @@ def execute(self, context):
         for prop in selection.__annotations__.keys():
             if hasattr(context.scene.dream_textures_prompt, prop):
                 setattr(context.scene.dream_textures_prompt, prop, getattr(selection, prop))
+            # when the seed of the promt is found in the available image datablocks, use that one in the open image editor
+            # note: when there is more than one image with the seed in it's name, do nothing. Same when no image with that seed is available.
+            if prop == 'seed':
+                seed_string = str(getattr(selection, prop))
+                image_names_no_suffix = [i.name.split('.')[0] for i in bpy.data.images]
+                image_names_no_suffix_seed_count = image_names_no_suffix.count(seed_string)
+                if image_names_no_suffix_seed_count == 1:
```
carson-katri, line 48: Does an exact match catch it correctly:
```suggestion
                if (existing_image := bpy.data.images.get(seed_string, None)) is not None:
```

operators/view_history.py
```diff
@@ -39,6 +39,22 @@ def execute(self, context):
...
+                    print("Image for seed " + seed_string + " is unique. Selecting image datablock.")
+                    for area in context.screen.areas:
+                        if area.type != 'IMAGE_EDITOR':
+                            continue
+                        area.spaces.active.image = bpy.data.images[seed_string]
```
carson-katri, line 53:
```suggestion
                        area.spaces.active.image = existing_image
                        self.report({'INFO'}, "Existing generated image opened in image editor")
```

operators/view_history.py
```diff
@@ -39,6 +39,22 @@ def execute(self, context):
...
+                elif image_names_no_suffix_seed_count > 1:
+                    self.report({'INFO'}, "More than one image found for seed " + seed_string + ". Not changing image editor contents.")
+                elif image_names_no_suffix_seed_count == 0:
+                    self.report({'INFO'}, "No image found for seed " + seed_string + ". Not changing image editor contents.")
```
carson-katri, line 57: I think I would rather show no message if an existing image isn't found, because users may think this is an error. I think it may be common to try reloading a prompt from a file that isn't open, so it may error more often than it doesn't.
```suggestion
```

operators/view_history.py
```diff
@@ -39,6 +39,22 @@ def execute(self, context):
...
             if hasattr(context.scene.dream_textures_prompt, prop):
                 setattr(context.scene.dream_textures_prompt, prop, getattr(selection, prop))
+            # when the seed of the promt is found in the available image datablocks, use that one in the open image editor
+            # note: when there is more than one image with the seed in it's name, do nothing. Same when no image with that seed is available.
+            if prop == 'seed':
+                seed_string = str(getattr(selection, prop))
+                image_names_no_suffix = [i.name.split('.')[0] for i in bpy.data.images]
+                image_names_no_suffix_seed_count = image_names_no_suffix.count(seed_string)
+                if image_names_no_suffix_seed_count == 1:
+                    print("Image for seed " + seed_string + " is unique. Selecting image datablock.")
```
carson-katri, line 49: Possibly remove the print statement to keep the logs clean.
```suggestion
```

carson-katri (changes requested):
Just a couple suggestions.

operators/view_history.py
GottfriedHofmann: Unfortunately the proposed change is not equivalent. Consider the following situation. We have made three runs with constant seed of 1096871657 because we wanted to see how the result looked with different samplers. Thus we have three images: 1096871657, 1096871657.001 and 1096871657.002.
Now I want to recall the second promt. With your change, it would select the first of the three images, which is the wrong one. Unfortunately we do not know which of the three is the right image with just the seed. That's why I have decided to count the occurences of a seed in the prefixes of the names of all image datablocks (line 46 and 47) and only select an image if it is the only one with that seed as prefix (line 48). 
When there is more than one occurence, it does not change the current image but reports that more than one has been found (lines 54 and 55).

I am aware that this is a heuristic and not perfect but I think I have an idea for a better solution. The problem is that the unique identifier for datablocks in Blender is the name, which is prone to be changed by the user and sideffects of Blender (.001 .002 .003 etc.). What we can do is the following: After an image is created, run hash() over the datablock and store the result as a custom property of said datablock and store it as well in dream_textures_prompt . That way we can definitely select the correct image no matter how it is named.

GottfriedHofmann:
I changed the way it works completely. Now a hash is created for each image and stored in a custom property of said image and in the promt history. When recalling a prompt from history and an image with said hash can be found in byp.data.images, show it in the image editor.

operators/dream_texture.py
```diff
@@ -96,8 +96,14 @@ def image_writer(shared_memory_name, seed, width, height, upscaled=False):
                     if area.type == 'IMAGE_EDITOR':
                         area.spaces.active.image = image
                 scene.dream_textures_prompt.seed = str(seed) # update property in case seed was sourced randomly or from hash
+                # create a hash from the Blender image datablock to use as unique ID of said image and store it in the prompt history
+                # and as custom property of the image. Needs to be a string because the int from the hash function is too large
+                image_hash = str(hash(image))
```
carson-katri, line 101: If you use hashlib to create a sha256 hash you can find the image even if it doesn't have the `dream_textures_hash` property, because it's consistent across runs.

```suggestion
                image_hash = hashlib.sha256((np.array(image.pixels) * 255).tobytes()).hexdigest()
```
Then compare against the hash again when reloading.

That may be too slow though, if so I'm ok merging as is.

GottfriedHofmann: If I understand correctly, the hash would need to be re-computed for on average half of the existing image datablocks when recalling the prompt. Doesn't that require lots of memory access?

carson-katri: I’m ok requiring the custom property on the image datablock. I’d still prefer using a sha256 hash of the image pixels because it will provide a consistent value in case we want to use it for other purposes, e.g. integration tests.

GottfriedHofmann: I'm agnostic to the hash method 🙂

carson-katri: My only issue with `hash(...)` is that it is not consistent across runs of Python afaik.

carson-katri (approved):
Looks good, thanks!

PR #405, opened 2022-11-28
🤗 Diffusers Backend

carson-katri:
This PR is large, but simplifies things quite a bit.

> I've noticed some weirdness using the Blender Development VS Code extension, if you start Blender and it throws an OSError, run `Reload Add-ons` and it should start up correctly. In a regular Blender run this doesn't seem to occur.

*I have not yet tested on Windows, so there are bound to be inconsistencies with multiprocessing and diffusers there.*

## Subprocess Updates

The generator process is now all based on `multiprocessing` and a custom `concurrent.futures.Future` subclass. The call site is simplified to:

```python
future = Generator.shared().prompt_to_image(prompt="cat", **other_args)
future.add_response_callback(on_progress) # Called when each response is received.
future.add_done_callback(on_done) # Called when the action exits
```

`Queue`s are used to pass messages back and forth. The `Generator` class is derived from `Actor`, which is the class the implements all of the logic for something loosely resembling the actor concurrency model.

All data is automatically pickled when sent through a queue, so it's really easy to pass inputs and outputs around.

Actions can be implemented on the generator as a simple function.

```python
# actions/custom_pipeline.py
def custom_pipeline(
  self,
  input_image: NDArray,
  steps: int
):
  # yield values, will call `on_response_callback`
  yield sampled_latents_1
  yield sampled_latents_2
  yield final_result

  # or use a return for actions that have no progress indicator.
  return 5

# __init__.py
class Generator(Actor):
  ...
  from .actions.custom_pipeline import custom_pipeline # This adds the method to the class

# other_file.py
future = Generator.shared().custom_pipeline(input_image, steps) # Automatically dispatched to a separate process and returns a future by the `Actor` base class.
```

## Pipelines
Three separate actions are provided:
* `prompt_to_image`
* `image_to_image`
* `inpaint`

These are implemented using 🧨 diffusers, and support Stable Diffusion v2.

## Upscaling Changes
Upscaling now uses StabilityAI's 4x upscaler instead of Real-ESRGAN. It is memory hungry, so tiling is implemented to split the generation up. This has a lot of room for improvement.

> The upscale model does not work correctly on MPS, so enable the "CPU Only" memory optimization under advanced for macOS.

## Model Management
Models can be downloaded from the Hugging Face Hub from the add-on preferences now.

## TODO
- [x] File names and history entries are incorrect
- [x] Upscaling seam blending
- [x] Generation progress bar

These could be tackled in separate PRs.

- [x] Re-add Stability SDK backend (#411)
- [ ] ONNX backend (for AMD users)
- [x] Outpainting (#412)
- [ ] Local checkpoint import and conversion - *required for release*


NullSenseStudio:
So far from what I see I really like the subprocess simplicity now.

> _I have not yet tested on Windows, so there are bound to be inconsistencies with multiprocessing and diffusers there._

I'm definitely getting issues there. Do you mind if I make changes directly on this branch or would you prefer me to leave those in review suggestions/separate PR?

> Upscaling now uses StabilityAI's 4x upscaler instead of Real-ESRGAN. It is memory hungry, so tiling is implemented to split the generation up. This has a lot of room for improvement.

I don't see any tile overlap and blending in the implementation. Does it cause any artifacts along tile edges?

> Models can be downloaded from the Hugging Face Hub from the add-on preferences now.

Does this entirely replace model importing? I expect people will still want to be able to load models that they download elsewhere.

carson-katri:
Feel free to work off this branch to fix things. If it’s a large change I’d prefer a separate PR though.

Overlap and blending still need to be implemented for upscaling.

Model importing needs to be added back in, but it will need to handle converting checkpoint files to diffusers’ format. This could be added in a separate PR.

NullSenseStudio:
Got it to a point where I can at least generate. Had to move dependencies loading and reorder `sys.path` to make sure `numpy` would be loaded from dependencies rather than from Blender. This newer version of `pytorch` was having an issue with that package.

Also noting that I've found more issues:
- installing dependencies from preferences is broken
- need to re-add platform specific `requirements.txt` for CUDA acceleration
- generate progress bar does not update
- downloaded model was not selectable from the dream panel until Blender was restarted
- attempting to download a model without an API key fills the preferences window with a large exception message, likely too confusing to end users
- downloading a model causes Blender to freeze for a while with the only progress indication being in an already open terminal, end users will likely believe it's broken

NullSenseStudio:
Speed on CUDA looks basically the same. VRAM usage is much higher. CUDA specific optimizations don't appear in in the dream UI, and the backend uses the same way to check if they're available so I don't think any of those are enabled. I'll have to check what's going on with that later.

carson-katri:
The main things to enable for VRAM usage would be fp16 (the model downloader may not be fetching the branches properly yet) and attention slicing.

NullSenseStudio:
Found and fixed what was causing those optimizations not being shown and used. Fixed the cancel message not being received on the backend, previously it would just pause due to the queue being full and resume once another action started handling the messages. Also fixed releasing the generator, every new actor was being instantiated with the same queues that were closed.

carson-katri:
With some of those optimizations enabled does the VRAM usage look more reasonable? We will need to pick good defaults so it works out of the box for most people.

NullSenseStudio:
I think the defaults are fine as they are. Only things I'd change are making inference mode always on and not even be an option, and move channels last format to be under speed optimizations.

generator_process/actor.py
```diff
@@ -38,6 +38,7 @@ def __init__(self):
         self._exception = None
         self.done = False
         self.cancelled = False
+        self.actor = actor
```
carson-katri, line 41: Where is the actor local variable it is assigning coming from?

NullSenseStudio: Forgot to remove it. I was thinking of having the future instantly send the cancel message instead of having it wait on the send loop. Current implementation will run another step before canceling but I don't think that's too big of an issue to worry about.

carson-katri: Ah ok. Yes, I'm fine with it being off a step.

carson-katri:
The core of this PR should be functional. Other additions can be made in separate PRs as noted in the initial comments.

generator_process/actions/huggingface_hub.py
```diff
@@ -0,0 +1,82 @@
...
+        force_download=False,
+        use_auth_token=token
+    )
+    # make sure we only download sub-folders and `diffusers` filenames
+    folder_names = [k for k in config_dict.keys() if not k.startswith("_")]
+    allow_patterns = [os.path.join(k, "*") for k in folder_names]
+    allow_patterns += [WEIGHTS_NAME, SCHEDULER_CONFIG_NAME, CONFIG_NAME, ONNX_WEIGHTS_NAME, StableDiffusionPipeline.config_name]
+
+    # make sure we don't download flax weights
+    ignore_patterns = "*.msgpack"
```
patrickvonplaten, line 66: Just a heads-up :-) Soon also safe tensors might be added to many diffusion repos since they are safe (compared to pickle) and fast. E.g. see: https://huggingface.co/CompVis/stable-diffusion-v1-4/discussions/171

So this might have to be added there if needed :-) 

The use case here is that you would like to download the model, but not load it into memory no? 
Think that's more or less exactly what @keturn was asking for as well

Maybe, it's time to better modularize `from_pretrained(...)` in `diffusers` to make this part of the code more robust :-)

generator_process/actions/image_to_image.py
```diff
@@ -0,0 +1,238 @@
...
+                    # corresponds to doing no classifier free guidance.
+                    do_classifier_free_guidance = guidance_scale > 1.0
+
+                    # 3. Encode input prompt
+                    text_embeddings = self._encode_prompt(
+                        prompt, device, num_images_per_prompt, do_classifier_free_guidance, negative_prompt
+                    )
+
+                    # 4. Preprocess image
+                    if isinstance(image, PIL.Image.Image):
```
patrickvonplaten, line 88: we very recently had some changes in `diffusers` to allow for batched image input - might be worth slightly adapting here :-)

generator_process/actions/image_to_image.py
```diff
@@ -0,0 +1,238 @@
...
+                            if do_classifier_free_guidance:
+                                noise_pred_uncond, noise_pred_text = noise_pred.chunk(2)
+                                noise_pred = noise_pred_uncond + guidance_scale * (noise_pred_text - noise_pred_uncond)
+
+                            # compute the previous noisy sample x_t -> x_t-1
+                            latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).prev_sample
+
+                            # NOTE: Modified to yield the latents instead of calling a callback.
+                            match kwargs['step_preview_mode']:
+                                case StepPreviewMode.NONE:
```
patrickvonplaten, line 125: very cool!

generator_process/actions/image_to_image.py
```diff
@@ -0,0 +1,238 @@
...
+                seed = random.randrange(0, np.iinfo(np.uint32).max)
+            generator = generator.manual_seed(seed)
+            
+            # Seamless
+            _configure_model_padding(pipe.unet, seamless, seamless_axes)
+            _configure_model_padding(pipe.vae, seamless, seamless_axes)
+
+            # Inference
+            with (torch.inference_mode() if device != 'mps' else nullcontext()), \
+                    (torch.autocast(device) if optimizations.can_use("amp", device) else nullcontext()):
```
patrickvonplaten, line 216: we quite strongly advise **against** using autocast since it's always slower than pure fp16

generator_process/actions/inpaint.py
```diff
@@ -0,0 +1,278 @@
...
+                    # corresponds to doing no classifier free guidance.
+                    do_classifier_free_guidance = guidance_scale > 1.0
+
+                    # 3. Encode input prompt
+                    text_embeddings = self._encode_prompt(
+                        prompt, device, num_images_per_prompt, do_classifier_free_guidance, negative_prompt
+                    )
+
+                    # 4. Preprocess mask and image
+                    if isinstance(image, PIL.Image.Image) and isinstance(mask_image, PIL.Image.Image):
```
patrickvonplaten, line 93: We also did some recent changes to allow batched image processing here :-)

patrickvonplaten (commented):
Very cool usage of `diffusers` - very interesting to see how it's used!

generator_process/actions/huggingface_hub.py
carson-katri: Yes, exactly. Something like `hf_hub_download` for `diffusers` models would be great!

generator_process/actions/image_to_image.py
carson-katri: Thanks!

generator_process/actions/image_to_image.py
carson-katri: Ah ok, does it provide any precision benefits?

carson-katri:
@patrickvonplaten Thanks for taking the time to check this out, `diffusers` has been really nice to work with so far!

generator_process/actions/image_to_image.py
patrickvonplaten: No, in our experience it doesn't. We recommend to pretty much always use fp16 when possible. Only if you would like max precision we recommend using full fp32 without autocast. 

We don't see any benefit in image quality when using autocast vs. fp16 and both autocast and fp16 seem to be a tiny tiny bit worse in terms of quality than fp32. So our recommendation is to either use pure fp16 or full fp32

PR #409, opened 2022-12-04
Add Project Dream Texture operator

carson-katri:
This uses depth to image to project a texture onto a mesh using a text prompt.


lawfordp2017:
Great idea but do you think thresholding the depth map at 0.5 limits what the depth map can do? I think the original model and the official diffusers pipeline use the full range of depth information, for instance nose may be closer to the camera than ears and so on.

carson-katri:
@lawfordp2017 Thanks for pointing that out, that was accidentally left in from the inpainting code

carson-katri:
Here's an example of generating a face texture onto a base mesh:

https://user-images.githubusercontent.com/13581484/206008139-cf98fc68-7066-4f6e-87a2-4c6e2acb1052.mp4

carson-katri:
Two more examples, using the Depth + Color option, and texturing multiple objects at once:

https://user-images.githubusercontent.com/13581484/206026218-aafb36a1-e067-48ea-b3fd-b42cd34415d2.mp4

https://user-images.githubusercontent.com/13581484/206026229-d631c05b-3260-4a9f-ac11-5a149926c3fa.mp4

Ryztiq:
I'm pretty new to Git, how would I go about using this in blender instead of the main branch?

carson-katri:
@Ryztiq Should be `git checkout sd-depth-projection`

ajwadimran7 (approved):

dddjef:
seems incredible!
is it compatible with 3.4? I've got "context is incorrect" error when I try to run "Project Dream Texture"

carson-katri:
@dddjef Make sure you enter edit mode and select all of the faces you want the texture to be applied to.

carson-katri:
I've added model type inference. It checks how many `in_channels` are specified in the U-Net config to guess what kind of model it is. This can be used to show hints before the user loads the wrong model that they need to switch to a depth-specific one or otherwise.

Screenshot 2022-12-11 at 3 58 41 PM

dddjef:
Tx, I don't have this error anymore, but it still doesn't work. Console shows
```
WARN (bpy.rna): C:\Users\blender\git\blender-v340\blender.git\source\blender\python\intern\bpy_rna.c:1341 pyrna_enum_to_py: current value '2' matches no enum in 'DreamPrompt', '', 'pipeline'
Traceback (most recent call last):
  File "C:\Users\dddje\AppData\Roaming\Blender Foundation\Blender\3.4\scripts\addons\dream_textures\operators\project.py", line 98, in draw
    row.enabled = Pipeline[context.scene.dream_textures_project_prompt.pipeline].depth()
  File "C:\Program Files\Blender Foundation\Blender 3.4\3.4\python\lib\enum.py", line 440, in __getitem__
    return cls._member_map_[name]
KeyError: ''
```
Not sure to install this correctly anyway. I've downloaded the big 7zip file from default dream texture repo. I unzip and install it, it works correctly. Then I set pycharm to the installed addon folder, and checkout your branch. All seems fine, I see your change, and the interface you added to the 3D view. Is is the right way to do it?

carson-katri:
@dddjef you would need to reinstall the dependencies in the `.python_dependencies` folder. In #405 everything was switched over to diffusers and new dependencies were added/removed.

FabianRueckert:
Using the latest commit, Blender doesn't find anything when I search for carsonkatri/stable-diffusion-2-depth-diffusers in the find model search box.
And the one from stabilityai directly doesn't work 
image

carson-katri:
@FabianRueckert thanks for giving it a try. Can you open the folder for that depth model and check the `snapshots/…/unet/config.json` file and see how many `in_channels` it has? It should be 5…
If so you could ignore the warning it should hypothetically work anyways. I’m not sure why my converted model doesn’t show in search.

dddjef:
> @dddjef you would need to reinstall the dependencies in the `.python_dependencies` folder. In #405 everything was switched over to diffusers and new dependencies were added/removed.

Sorry, I still struggle with this. I removed everything from .python_dependencies folder, and then I use pip install with requirements/win-linux-cuda.txt but the process stops with an error  `Could not find a version that satisfies the requirement torch>=1.13 (from versions: none)`

FabianRueckert:
@carson-katri Ok I got it to work. I think the reason for the error above was, that the model wasn't fully downloaded due to a choppy connection,but the script thought it had been. Deleting it and redownloading worked.

NullSenseStudio (approved):
LGTM

PR #410, opened 2022-12-05
Automatic Seamless Detection

NullSenseStudio:
Adds automatic seamless axis detection to simplify source image and upscaling (when supported) workflows. Currently only shows the detected result, but once finished it'll have options to automatically apply the result, prompt the user to change, or have detection disabled.

blender_oORSDXZFIt

Detection is done through a custom trained model on 2000 generated images with 2x and 4x Real-ESRGAN upscaled versions of each. Accuracy is 99.75% per edge or 99.5% per image and may vary somewhat depending on the type of textures that're run through it.


carson-katri:
Looks great! Would you want to include the training script (possibly with a top level `scripts` folder)?

My suggestion for the interface would be to change the "Seamless" checkbox to a dropdown with the options:
* Automatic
* X
* Y
* Both
* Off

where "Automatic" is the default (or maybe "Auto-detect"). What do you think?

NullSenseStudio:
I plan on adding the training script but it's quite a mess right now and really should have comments explaining certain parts of it.

I think naming the option "Auto-detect" would be suitable.

NullSenseStudio:
Seamless axes are now handled as a single enum property with the detected result appearing below, and will only run if source image is checked and has modify or inpaint selected. It can also now look through prompt history for how an image was generated and fallback to the trained model if nothing was found.
blender_tTPsTDPqnv
It's also possible to use with texture projection and the render pass, but the user won't be notified of what seamless mode it's using. It might be better to change the default to off instead of auto-detect as I don't see it being likely to be used outside of modifying or inpainting.

generator_process/actions/detect_seamless/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+        model = SeamlessModel()
+        model.load_state_dict(state)
+        model.eval()
+        setattr(self, 'detect_seamless_model', model)
+
+    if torch.cuda.is_available():
+        device = 'cuda'
+    elif torch.backends.mps.is_available():
+        device = 'mps'
```
carson-katri, line 119: The model does not output correct results on `mps`. I poked around a bit and it seems it seems to deviate from the CPU results most through the `GRU` layer, but I think using CPU is fine as it seems to run just as fast.
```suggestion
        device = 'cpu'
```

carson-katri (commented):
Looks great! Just a couple suggestions.

NullSenseStudio:
Do you normally squash and merge when complete?

carson-katri:
Yep

PR #412, opened 2022-12-06
Outpainting for #405

carson-katri:
I changed the way outpainting works for better performance and flexibility. Instead of specifying the expansion on each axis, it uses the specified image size and an origin point.

For example, to extend a 512x512 image to the right:
1. Set an image size to specify the outpaint cursor area (e.g. `(512, 512)`)
2. Set the origin point to `(448, 0)` for a 64px overlap, `(384, 0) for a 128px overlap, etc.

It will only run SD on a 512x512 area, and then overlays it on an extended image at the end.

> An inpainting finetuned model must be used for outpainting.


generator_process/actions/outpaint.py
```diff
@@ -0,0 +1,74 @@
...
+    if outpaint_origin[0] > init_image.size[0] or outpaint_origin[0] < -width:
+        raise ValueError(f"Outpaint origin X ({outpaint_origin[0]}) must be between {-width} and {init_image.size[0]}")
+    if outpaint_origin[1] > init_image.size[1] or outpaint_origin[1] < -height:
+        raise ValueError(f"Outpaint origin Y ({outpaint_origin[1]}) must be between {-height} and {init_image.size[1]}")
+    
+    outpaint_bounds = Image.new(
+        'RGBA',
+        (
+            max(init_image.size[0], outpaint_origin[0] + width),
+            max(init_image.size[1], outpaint_origin[1] + height),
```
NullSenseStudio, line 31: Image gets cropped when outpainting left or up.
```suggestion
            max(init_image.size[0], outpaint_origin[0] + width) - min(0, outpaint_origin[0]),
            max(init_image.size[1], outpaint_origin[1] + height) - min(0, outpaint_origin[1]),
```

NullSenseStudio (changes requested):
So far mostly works great, some issues though.

When the outpainted image gets large enough I'm receiving this error right after the final image gets placed in Blender. I'm assuming it has to do with it being done in a background thread and not Blender's main thread.
```
  File "...\dream-textures\operators\dream_texture.py", line 147, in done_callback
    image['dream_textures_hash'] = image_hash
AttributeError: Writing to ID classes in this context is not allowed: 388423733, Image datablock, error setting Image.dream_textures_hash
```

Also for some reason there is color shifting occurring in parts of the image that aren't even being outpainted to at the time.
blender_4hO1CJdKoK

carson-katri:
I haven't had it happen yet with this implementation, but it seemed pretty random in the first place so hopefully this really did fix it.

One drawback of this approach is it removes the ability to block via the `Future`:
```python
Generator.shared().prompt_to_image(...).result()
```
Instead you need the more verbose
```python
Generator.shared().prompt_to_image(..., _block=True).result()
```
I don't think this is too bad considering it's internal API.
And perhaps we shouldn't block either way for UX.

NullSenseStudio:
Perhaps the `_send()` function could remain how it was. Have a single threading event for each future that can be waited on for `result()` rather than adding in another callback. Then only create app timers when callbacks are run.

carson-katri:
I think 9480c4b works the way you described, which does seem a bit better than the artificial delays that come with using timers as a thread.

preferences.py
```diff
@@ -156,7 +156,7 @@ class StableDiffusionPreferences(bpy.types.AddonPreferences):
     @staticmethod
     def register():
         if Pipeline.local_available():
-            set_model_list('installed_models', Generator.shared().hf_list_installed_models().result())
+            set_model_list('installed_models', Generator.shared().hf_list_installed_models(_block=True).result())
```
NullSenseStudio, line 159: Attempting to start Blender via the VS Code extension is currently freezing. Changing this will allow it to start for me but requires disabling and re-enabling the addon for anything that runs on the backend. Has a downside of the models not being available right away but, also doesn't slow down Blender since it doesn't have to wait on the backend to start.
```suggestion
            def on_done(future):
                set_model_list('installed_models', future.result())
            Generator.shared().hf_list_installed_models().add_done_callback(on_done)
```

carson-katri: I think removing the delay from Blender's launch time is better anyways.

NullSenseStudio:
> Also for some reason there is color shifting occurring in parts of the image that aren't even being outpainted to at the time.

I've tracked this down to being caused by `save_temp_image()`. Had an image generated, made a copy via that function and brought it back into Blender with the same color shifting being visible. Perhaps it could be sent to the backend as an ndarray instead of making a temp file?

carson-katri:
<table>
<tr>
<td valign="bottom">



</td>
<td>



</td>
</tr>
</table>

NullSenseStudio (approved):
Was just about to mention those axis needing switched. Really getting tired of whoever standardized the y axis going before x.

Anyhow, everything seems to be working as intended now!

carson-katri:
Me too, it's bizarre. They put the `height` before the `width` argument in the pipelines too.

PR #462, opened 2022-12-21
add prompt string to material and texture names

JasonHoku:
Minor feature, improves organization.

I'm interested to hear if this has any cons.

Thank you


carson-katri:
I do think this could be valuable, and I like how it only adds the subject. Perhaps it could be formatted like so instead:

```
Prompt Subject (Seed)
```

I don't mind spaces in the image name so you can keep them instead of replacing with `-`.

Also, can you target the `main` branch instead of `detect-seamless`.

carson-katri:
It looks like you may have rebased so every commit in the project to shows up here. Please fix this if you can.

PR #479, opened 2022-12-25
Properly handle `fp16` branches

carson-katri:
Fixes an issue where errors would appear trying to load a model that only had a `fp16` revision downloaded. This could happen if a user previously used `diffusers` to download a model outside of the add-on.

A new checkbox is also added to the "Find Models" box to prefer downloading `fp16` branches instead of `main` if possible. This is very useful for users who only run in "Half Precision" and don't need to download the larger weights. If the model doesn't have a `fp16` branch, it falls back on whichever branch is default.


generator_process/actions/prompt_to_image.py
```diff
@@ -236,6 +236,25 @@ def approximate_decoded_latents(latents):
...
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            if revision == preferred_revision:
+                break
+        if ref_path is None:
+            return None
+        with open(ref_path) as f:
+            commit_hash = f.read()
+
+        snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)
+    return snapshot_folder
```
NullSenseStudio, line 256:
```suggestion
        if len(os.listdir(snapshot_folder)) <= 1:
            raise ValueError(f"revision \"{revision}\" is not installed for model {model}")
    return snapshot_folder
```
Downloading the fp16 branch also downloads `model_index.json` for the main branch. This leads to the function returning the partially downloaded main snapshot when someone tries to disable half precision while using the model. It'll later error with `OSError: Error no file named diffusion_pytorch_model.bin` which is ambiguous with an incorrectly downloaded model.

generator_process/actions/prompt_to_image.py
```diff
@@ -236,6 +236,25 @@ def approximate_decoded_latents(latents):
...
     return latents_ubyte.numpy()
 
+def model_snapshot_folder(model, preferred_revision: str | None = None):
+    """ Try to find the preferred revision, but fallback to another revision if necessary. """
+    storage_folder = model
+    if os.path.exists(os.path.join(storage_folder, 'model_index.json')): # converted model
+        snapshot_folder = storage_folder
+    else: # hub model
+        ref_path = None
+        for revision in os.listdir(os.path.join(storage_folder, "refs")):
```
NullSenseStudio, line 246: When `preferred_revision` is None it's expecting a fp32 model, but this will lead to using whichever is the last revision depending on how the os returns files from listdir(). This may lead to incorrect behavior when multiple revisions are downloaded.

NullSenseStudio (changes requested):
Mostly good. There's a few issues however.

In hf_snapshot_download()
`from diffusers.hub_utils import http_user_agent` needs changed to
`from diffusers.utils.hub_utils import http_user_agent` due to a recent change in diffusers.

generator_process/actions/prompt_to_image.py
carson-katri: I think loading the `fp16` revision should still work, even if half precision is disabled.

generator_process/actions/huggingface_hub.py
```diff
@@ -109,12 +124,12 @@ def hf_snapshot_download(
...
     from huggingface_hub.hf_api import HfApi
-    from huggingface_hub.utils import filter_repo_objects, validate_hf_hub_args, tqdm, logging, EntryNotFoundError, LocalEntryNotFoundError
+    from huggingface_hub.utils import filter_repo_objects, validate_hf_hub_args, tqdm, logging, EntryNotFoundError, LocalEntryNotFoundError, RevisionNotFoundError
 
     from diffusers import StableDiffusionPipeline
     from diffusers.utils import DIFFUSERS_CACHE, WEIGHTS_NAME, CONFIG_NAME, ONNX_WEIGHTS_NAME
     from diffusers.schedulers.scheduling_utils import SCHEDULER_CONFIG_NAME
-    from diffusers.hub_utils import http_user_agent
+    from diffusers.utils.hub_utils import http_user_agent
     config_dict = StableDiffusionPipeline.get_config_dict(
```
NullSenseStudio, line 133: FutureWarning:  The function get_config_dict is deprecated. Please use <class '...StableDiffusionPipeline'>.load_config instead. This function will be removed in version v1.0.0
```suggestion
    config_dict = StableDiffusionPipeline.load_config(
```

NullSenseStudio (changes requested):
All good besides a deprecation warning I noticed.

PR #480, opened 2022-12-26
Stop reusing the first UV map slot for texture projection

Dekker3D:
This change makes texture projection no longer reuse the first UV map in an object. Instead, it will always use a map with the name "Projected UVs", and add a UV Map node to the node tree to refer to this map.


carson-katri (approved):
LGTM 👍

PR #485, opened 2022-12-28
Fix Iterations and File Batch

NullSenseStudio:
Prevents the resulting images from being overwritten when using iterations or file batch. History entries are now only created once an image is finished to fix a crash that can be caused from creating and keeping references to multiple entries at a time, and will no longer leave extra duplicate entries if generation is stopped early. Also fixes a bug where generation would continue to the next iteration when attempting to cancel.


carson-katri (approved):
👍 LGTM

PR #495, opened 2023-01-02
Automatic Projection Baking

carson-katri:
This adds an option to automatically bake the generated texture to a different UV map. It should only bake the selected faces, so a future PR can merge multiple baked generations together and blend the seams.

I also re-implemented the depth rendering to use `GPUOffScreen` instead of adding a callback and updating.

> **Note**: You need to be in face select mode with the target faces selected so it can identify which faces need to be baked.

Screenshot_2023-01-01_at_6 50 30_PM
Screenshot 2023-01-01 at 6 53 38 PM
Screenshot_2023-01-01_at_6 50 44_PM


operators/project.py
```diff
@@ -119,71 +128,93 @@ def draw(self, context):
...
+            vertices = np.array([[l.vert.index for l in loop] for loop in mesh.calc_loop_triangles()], dtype='i')
+
+            shader = bake_shader()
+            batch = batch_for_shader(
+                shader, 'TRIS',
+                {"src_uv": src_uv, "dest_uv": dest_uv},
+                indices=vertices,
+            )
+            shader.uniform_sampler("image", texture)
+            batch.draw(shader)
```
NullSenseStudio, line 214: 'LESS_EQUAL' and 'LESS' depth_test modes don't appear to work at all on my system, and none of the rest are getting any better result than coloring all of the selected UV zones a single color.

The viewport freezes and other spaces had issues redrawing when depth_test is left as 'LESS_EQUAL'
```suggestion
        prev_depth_test = gpu.state.depth_test_get()
        prev_depth_mask = gpu.state.depth_mask_get()
        gpu.state.depth_test_set('LESS_EQUAL')
        gpu.state.depth_mask_set(True)
        with gpu.matrix.push_pop():
            gpu.matrix.load_matrix(mathutils.Matrix.Identity(4))
            gpu.matrix.load_projection_matrix(mathutils.Matrix.Identity(4))

            vertices = np.array([[l.vert.index for l in loop] for loop in mesh.calc_loop_triangles()], dtype='i')

            shader = bake_shader()
            batch = batch_for_shader(
                shader, 'TRIS',
                {"src_uv": src_uv, "dest_uv": dest_uv},
                indices=vertices,
            )
            shader.uniform_sampler("image", texture)
            batch.draw(shader)
        gpu.state.depth_test_set(prev_depth_test)
        gpu.state.depth_mask_set(prev_depth_mask)
```

NullSenseStudio (approved):
That actually worked. I wish it occurred to me to try removing those lines instead of just changing their parameters.

carson-katri:
I changed the default prompt structure to 'Custom', as many users default leave the 'Texture' preset selected when they aren't generating textures.

PR #499, opened 2023-01-03
Add Prompt String To Material And Texture Names

JasonHoku:


carson-katri:
There are some conflicts now, could you resolve those?

JasonHoku:
> There are some conflicts now, could you resolve those?

Resolved and tested 👍

JasonHoku:
Oops, looks like my second PR got merged into this branch. I believe I've corrected it though and will make a second PR on another branch for the readme troubleshooting addition.

JasonHoku:
I found that the naming in dream_textures.py was not working properly, cleaned up and fixed in c3ba80d.

operators/dream_texture.py
```diff
@@ -105,7 +105,9 @@ def done_callback(future):
             result: ImageGenerationResult = future.result(last_only=True)
             for i, result_image in enumerate(result.images):
                 seed = result.seeds[i]
-                image = bpy_image(str(seed), result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
+                prompt_string = bpy.data.scenes["Scene"].dream_textures_prompt.prompt_structure_token_subject
+
+                image = bpy_image(str(prompt_string + " " + str(seed)), result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```
carson-katri, line 110: I think an f-string would be a bit simpler here. And putting the seed in parentheses or after a dash may be easier to read:

```suggestion
                image = bpy_image(f"{prompt_string} ({seed})", result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```

operators/project.py
```diff
@@ -354,9 +354,15 @@ def on_done(future):
...
-                texture = bpy.data.images.new(name=str(generated.seeds[0]), width=generated.images[0].shape[1], height=generated.images[0].shape[0])
-            texture.name = str(generated.seeds[0])
-            material.name = str(generated.seeds[0])
+                gen_seed_string = str(generated.seeds[0])
+                prompt_subject = bpy.data.scenes["Scene"].dream_textures_project_prompt.prompt_structure_token_subject
+                name_with_prompt = str(prompt_subject + " " + gen_seed_string)
+                texture = bpy.data.images.new(name=name_with_prompt, width=generated.images[0].shape[1], height=generated.images[0].shape[0])
+            gen_seed_string = str(generated.seeds[0])
+            prompt_subject = bpy.data.scenes["Scene"].dream_textures_project_prompt.prompt_structure_token_subject
+            name_with_prompt = str(prompt_subject + " " + gen_seed_string)
```
carson-katri, line 363: Seems to be some duplication here, could this section be moved above the `if` so it only has to be written once? And an f-string could be simpler here too.

```python
gen_seed_string = str(generated.seeds[0])
prompt_subject = bpy.data.scenes["Scene"].dream_textures_project_prompt.prompt_structure_token_subject
name_with_prompt = f"{prompt_subject} ({gen_seed_string})"
```

carson-katri (changes requested):
This will be a nice usability improvement, just a couple small suggestions.

operators/project.py
```diff
@@ -354,9 +354,15 @@ def on_done(future):
...
             if texture is None:
-                texture = bpy.data.images.new(name=str(generated.seeds[0]), width=generated.images[0].shape[1], height=generated.images[0].shape[0])
-            texture.name = str(generated.seeds[0])
-            material.name = str(generated.seeds[0])
+                gen_seed_string = str(generated.seeds[0])
+                prompt_subject = bpy.data.scenes["Scene"].dream_textures_project_prompt.prompt_structure_token_subject
+                name_with_prompt = str(prompt_subject + " " + gen_seed_string)
+                texture = bpy.data.images.new(name=name_with_prompt, width=generated.images[0].shape[1], height=generated.images[0].shape[0])
+            gen_seed_string = str(generated.seeds[0])
+            prompt_subject = bpy.data.scenes["Scene"].dream_textures_project_prompt.prompt_structure_token_subject
```
carson-katri, line 362: Can you access the scene from the context instead of by name? I think this will fail if the scene is renamed.

JasonHoku: Resolved in [ca964e3](https://github.com/carson-katri/dream-textures/pull/499/commits/ca964e314dd003ef8318a7838647778a5d406526)

operators/dream_texture.py
```diff
@@ -105,7 +105,9 @@ def done_callback(future):
             result: ImageGenerationResult = future.result(last_only=True)
             for i, result_image in enumerate(result.images):
                 seed = result.seeds[i]
-                image = bpy_image(str(seed), result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
+                prompt_string = context.scene.dream_textures_prompt.prompt_structure_token_subject
+
+                image = bpy_image(f"{prompt_string} ({seed})", result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```
carson-katri, line 110: Blender truncates image names to 63 characters, which means if the prompt is long the seed will be dropped automatically. Two possible solutions:
* Swap so the prompt is in parentheses after the seed
* Check the length of the seed and trim the `prompt_string` accordingly so it all fits - If Blender increases this limit in the future though we'd have to remember to update on our end
```suggestion
                image = bpy_image(f"{seed} ({prompt_string})", result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```

operators/project.py
```diff
@@ -351,12 +351,15 @@ def on_done(future):
             context.scene.dream_textures_info = ""
             context.scene.dream_textures_progress = 0
             generated = future.result()
+            gen_seed_string = str(generated.seeds[0])
+            prompt_subject = context.scene.dream_textures_project_prompt.prompt_structure_token_subject
+            name_with_prompt = str(prompt_subject + " " + gen_seed_string)
```
carson-katri, line 356: Can we use an f-string here too?

> I included the first option of the above comment in this suggestion, but if you think the seed on the end is better that would work too

```suggestion
            prompt_subject = context.scene.dream_textures_project_prompt.prompt_structure_token_subject
            name_with_prompt = f"{generated.seeds[0]} ({prompt_subject})"
```

carson-katri (changes requested):
Noticed an issue with long prompts. I had two ideas but maybe you have a better solution.

operators/dream_texture.py
JasonHoku: The only issue I have with putting the seed first is that we lose the default alphabetic sorting within blender and it becomes less organized when selecting images in the Image Editor.

One solution to that could be to use a counter, like how the Stable Diffusion WebUI does it, which would then chronologically order the outputs. Another benefit to this is that you will not overwrite duplicate named images.

Otherwise I think for now trimming the names length to 63 minus the seeds string length seems like a good solution.

JasonHoku:
```
seed_str_length = len(str(seed))
trim_aware_name = (prompt_string[:58 - seed_str_length] + '..') if len(prompt_string) > 58 else prompt_string
name_with_trimmed_prompt = f"{trim_aware_name} ({seed})"
```
image

I've come up with this which seems to work quite nicely. Commit incoming.

JasonHoku:
I found a weird discrepancy in project.py that causes the seed to be accessed differently than in dream_texture.py:

dream_texture.py: `seed = result.seeds[i]`

project.py: `seed = generated[0].seeds[0]`

Though I've named everything accordingly and tested them in [2c1623f](https://github.com/carson-katri/dream-textures/pull/499/commits/2c1623f01bcba22441bd278018a00f43bbf6448e)

That aught to do it, lmk what you think! ^_^

carson-katri:
I think that's because `dream_texture` has the `iterations` option to generate multiple images but `project` can only generate one.

operators/dream_texture.py
```diff
@@ -105,7 +105,11 @@ def done_callback(future):
             result: ImageGenerationResult = future.result(last_only=True)
             for i, result_image in enumerate(result.images):
                 seed = result.seeds[i]
-                image = bpy_image(str(seed), result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
+                prompt_string = context.scene.dream_textures_prompt.prompt_structure_token_subject
+                seed_str_length = len(str(seed))
+                trim_aware_name = (prompt_string[:58 - seed_str_length] + '..') if len(prompt_string) > 58 else prompt_string
+                name_with_trimmed_prompt = f"{trim_aware_name} ({seed})"
+                image = bpy_image(f"{name_with_trimmed_prompt} ({seed})", result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```
carson-katri, line 112: Looks like this adds the seed to the name twice, but you can't see it in the result because it gets truncated 🙃 

```suggestion
                image = bpy_image(name_with_trimmed_prompt, result_image.shape[1], result_image.shape[0], result_image.ravel(), last_data_block)
```

JasonHoku:
btw I believe I noticed that Blender automatically appends a numerical counter to duplicately named images which adds four more characters to the name like `.001` and so will trim out part of the seed in those cases. 

So maybe we should change the 58 character count to 54 to remedy this.

JasonHoku:
That seems to do it!

image

Though I never noticed this error before, got a bad output on every other render (weird) and had to pull up the console to find this message:
`The following part of your input was truncated because CLIP can only handle sequences up to 77 tokens: ['2 3 4 5 6 7 8 9']`

carson-katri (approved):
Works great, thanks!

JasonHoku:
Woohoo! Though I might mention that it seems even though blender appends the counter to duplicately named images it still will export/save them without the counter and can cause some overwrite issues.

If we added our own three digit counter it would replace the Blender duplicate counter and alleviate the export issue but that might be more fringe of an fix/enhancement for later down the road.

carson-katri:
That seems like something Blender would need to change or provide some option for.

PR #553, opened 2023-02-05
Force face selection mode

carson-katri:
Resolves an issue when using the "Bake" option without being in face selection mode.


NullSenseStudio (approved):
LGTM

PR #584, opened 2023-03-02
Improvement: Add shortcuts PAGE_UP and PAGE_DOWN in the image editor …

GottfriedHofmann:
…to switch to the next or previous image. Created using ChatGPT in Bing. I thought using ChatGPT for this is fun and somewhat meta, I hope it is also ok.

The keys are currently hardcoded to PAGE_UP and PAGE_DOWN. I don't think init.py is the right place for this but rather somewhere else in the code. No need to put the keys into the addon preferences, though because the user is able to change the keys using the Blender keymap editor once the addon is registered:

image

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -16,7 +16,7 @@ bl_info = {
     "author": "Dream Textures contributors",
     "description": "Use Stable Diffusion to generate unique textures straight from the shader editor.",
     "blender": (3, 1, 0),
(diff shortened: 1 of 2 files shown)


carson-katri:
To be honest, this seems like something that would make more sense as either an addition to Blender itself, or a separate add-on to add more key bindings to Blender. Simply because this is only tangentially related to AI image generation.

PR #614, opened 2023-03-25
Add segmentation ControlNet support

carson-katri:
This adds built-in support for the segmentation ControlNet model. Apply labels to objects then use them in the render engine to produce segmentation maps that use the ADE20K classes/colors.

segmentation
donut
Screenshot 2023-03-20 at 6 03 59 PM

### DIFF
--- a/engine/annotations/ade20k.py
+++ b/engine/annotations/ade20k.py
@@ -69,6 +69,7 @@ def render_ade20k_map(context, collection=None, invert=True):
                     draw_annotation(vertices, indices, color)
             result = np.array(fb.read_color(0, 0, width, height, 4, 0, 'FLOAT').to_list())
             result[:, :, 3] = 1
+        gpu.state.depth_test_set('NONE')
         offscreen.free()
         e.set()
     bpy.app.timers.register(_execute, first_interval=0)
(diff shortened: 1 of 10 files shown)


engine/annotations/ade20k.py
```diff
@@ -0,0 +1,99 @@
...
+    return result
+
+def draw_annotation(vertices, indices, color):
+    shader = gpu.shader.from_builtin('3D_UNIFORM_COLOR')
+    batch = batch_for_shader(
+        shader, 'TRIS',
+        {"pos": vertices},
+        indices=indices,
+    )
+    shader.uniform_float("color", color)
```
NullSenseStudio, line 98: I've run into the same color bug that was previously encountered for the openpose annotation.
[Bind the shader object. Required to be able to change uniforms of this shader.](https://docs.blender.org/api/current/gpu.types.html#gpu.types.GPUShader.bind)
Turns out that `batch.draw(shader)` would bind the shader and allow the color to be properly set after. This can be similarly changed for openpose instead of using the twice drawn background workaround.
```suggestion
    shader.bind()
    shader.uniform_float("color", color)
```

NullSenseStudio (changes requested):
Looking good. Just run into some minor issues with OpenGL rendering again

PR #616, opened 2023-03-27
ControlNet + Inpainting Projection

carson-katri:
## Usage
Improved UI still needs to be created. For now:

1. Place cameras around a single object
2. Select the object and go into edit mode
3. In the projection panel, enable "Use ControlNet" and choose a depth ControlNet model
4. Ensure you have downloaded the models `runwayml/stable-diffusion-v1-5` and `runwayml/stable-diffusion-inpainting` as the model is currently hard-coded
5. Press "Project", Blender will hang while the generation occurs

### For the first perspective

1. Generate the first angle with a depth ControlNet and text to image model.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L528-L537

generation_result

2. Bake the texture to the result.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L538

### For subsequent perspectives

1. Create a mask for faces oriented toward this camera, in both projected and baked forms.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L353

mask

2. Capture the object with the latest result texture from this new perspective, and use the mask as the alpha channel.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L510-L511

3. Inpaint using the masked `color` with a depth ControlNet and inpainting model.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L512-L523

inpaint_result

7. Bake and merge with the latest result texture.

https://github.com/carson-katri/dream-textures/blob/7884149f6cc09c820d4e16dd7964af6004e149d0/operators/project.py#L524-L526

## Results
I have not thoroughly tested this, but here is one result using only two camera angles.

Screenshot 2023-03-27 at 7 21 58 PM

Coherence between the two sides is better than if two completely random generations were merged.

### DIFF
--- a/generator_process/actions/control_net.py
+++ b/generator_process/actions/control_net.py
@@ -23,6 +23,11 @@ def control_net(
     controlnet_conditioning_scale: list[float],
     
     image: NDArray | str | None, # image to image
+    # inpaint
+    inpaint: bool,
+    inpaint_mask_src: str,
+    text_mask: str,
+    text_mask_confidence: float,
 
     strength: float,
     prompt: str | list[str],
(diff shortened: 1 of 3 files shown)


kamtar:
I might need to try to finish it, but python and blender plugins isn't something I have done before.

If you have some pointers about what exactly needs finishing I would be glad. Are UI and hard-coded models the only things missing?

PR #625, opened 2023-04-07
Support Inpainting with ControlNet

carson-katri:
Fairly simple implementation. The inpainting-specific latents are created as-needed in the `control_net` action. Other inpainting mask logic is copied from the `inpaint` action.


NullSenseStudio (approved):
LGTM

PR #626, opened 2023-04-07
Add more nodes

carson-katri:
This adds some nodes that are useful for things like feedback loops and image manipulation.

New nodes:
* Image File - opens an image from a file path (Blender 3.5+ due to OpenImageIO availability)
* Frame Path - Gets the path to a frame at a specific index
* Crop Image - Crops an image with a bounding rect
* Join Images - Stacks two images horizontally or vertically
* Separate Color - Splits image into RGBA channels
* Combine Color - Creates image from RGBA channels
* Switch - If statement
* Compare - `>`, `<`, `==`, etc.
* Replace String - Find and replace in string


antigo333:
Add a Mix Node. (+FactorSocked)
The mix node with its standard image operations is the core element of any image processing and should therefore be implemented at an early stage. 
The second important node would be Transform/Scale.

PR #632, opened 2023-04-14
Add normal map generation

carson-katri:
This adds support for generating normal maps compatible with the [`control_v11p_sd15_normalbae`](https://huggingface.co/lllyasviel/control_v11p_sd15_normalbae) model.
sample


carson-katri:
Some edge cases were found, but I think they can be tackled in separate PRs. I'd like to get the ControlNet stuff released and refine from there.

alanhzh:
How to use normal map ? Is there any tutorials?

PR #636, opened 2023-04-17
Bump Torch ROCm version to 2.0

cornpo:
Currently we will miss ROCm wheel and install Torch-CUDA with linux-rocm.txt line "torch>=1.13" so I bumped the --extra-index-url to the new ROCm and the version to 2.0


carson-katri (approved):
Nice, thanks!

PR #646, opened 2023-04-25
Hide NodeResizeImage and NodeImageFile if not v3.5

phi-line:
These files were not registered, but they were still present in the node categories. This commit follows the same hiding logic as in (un)registration.

This was my menu in 3.4, note the "Unknown" node
image

I since upgraded to 3.5, but I wanted to fix this for other 3.4 users!

### DIFF
--- a/engine/__init__.py
+++ b/engine/__init__.py
@@ -17,45 +17,57 @@ class DreamTexturesNodeCategory(nodeitems_utils.NodeCategory):
     def poll(cls, context):
         return context.space_data.tree_type == DreamTexturesNodeTree.__name__
 
+pipeline_items = [
+    nodeitems_utils.NodeItem(NodeStableDiffusion.bl_idname),
+    nodeitems_utils.NodeItem(NodeControlNet.bl_idname),
+]
+
+input_items = [
+    nodeitems_utils.NodeItem(NodeInteger.bl_idname),
+    nodeitems_utils.NodeItem(NodeString.bl_idname),
+    nodeitems_utils.NodeItem(NodeImage.bl_idname),
+    nodeitems_utils.NodeItem(NodeCollection.bl_idname),
+    nodeitems_utils.NodeItem(NodeRenderProperties.bl_idname),
+]
+if bpy.app.version >= (3, 5, 0):
+    input_items.append(nodeitems_utils.NodeItem(NodeImageFile.bl_idname))
+
+utility_items = [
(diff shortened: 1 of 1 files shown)


engine/__init__.py
```diff
@@ -17,45 +17,58 @@ class DreamTexturesNodeCategory(nodeitems_utils.NodeCategory):
...
+    nodeitems_utils.NodeItem(NodeSwitch.bl_idname),
+    nodeitems_utils.NodeItem(NodeCompare.bl_idname),
+    nodeitems_utils.NodeItem(NodeReplaceString.bl_idname),
+]
+if bpy.app.version >= (3, 5, 0):
+    utility_items.append(nodeitems_utils.NodeItem(NodeResizeImage.bl_idname))
+
+annotations_items = [
+    nodeitems_utils.NodeItem(NodeAnnotationDepth.bl_idname),
+    nodeitems_utils.NodeItem(NodeAnnotationNormal.bl_idname),
```
phi-line, line 56: @carson-katri This is from your changes that are on `main` https://github.com/carson-katri/dream-textures/commit/66a148518eb8a697e1b8b4e3e08937b20cf78ce5. We can remove this from this PR if you'd like, but it will need to be added back in anyways before this lands on `main`

carson-katri: I'd prefer this to be removed, I'll rebase the other branch.

engine/__init__.py
```diff
@@ -17,45 +17,58 @@ class DreamTexturesNodeCategory(nodeitems_utils.NodeCategory):
...
+    nodeitems_utils.NodeItem(NodeInteger.bl_idname),
+    nodeitems_utils.NodeItem(NodeString.bl_idname),
+    nodeitems_utils.NodeItem(NodeImage.bl_idname),
+    nodeitems_utils.NodeItem(NodeCollection.bl_idname),
+    nodeitems_utils.NodeItem(NodeRenderProperties.bl_idname),
+]
+if bpy.app.version >= (3, 5, 0):
+    input_items.append(nodeitems_utils.NodeItem(NodeImageFile.bl_idname))
+
+utility_items =[
```
carson-katri, line 35:
```suggestion
utility_items = [
```

phi-line:
@carson-katri I addressed your comments. This should be good to go

aulerius:
On 3.5.1 I do not see several image related nodes, could that be related to this?

PR #659, opened 2023-05-06
Various UI Tweaks

NullSenseStudio:
- When there are multiple image editors, generating from an open image source or upscaling will now pick the image from the image editor it was executed from.
- Image editors with image pin activated will no longer be replaced with generated images.
- Cancel generator operator has been slightly changed to better differentiate it from release generator.
blender_nFIlZtKP8q
- Image texture nodes will now fill the shader editor view better instead of just the bottom right corner.
blender_NpISzg732d

### DIFF
--- a/operators/dream_texture.py
+++ b/operators/dream_texture.py
@@ -2,6 +2,7 @@ import bpy
 import hashlib
 import numpy as np
 import math
+from typing import Literal
 
 from .notify_result import NotifyResult
 from ..preferences import StableDiffusionPreferences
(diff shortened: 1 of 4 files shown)


operators/dream_texture.py
```diff
@@ -80,13 +92,7 @@ def execute(self, context):
             )
 
         # Setup the progress indicator
-        def step_progress_update(self, context):
-            if hasattr(context.area, "regions"):
-                for region in context.area.regions:
-                    if region.type == "UI":
-                        region.tag_redraw()
-            return None
-        bpy.types.Scene.dream_textures_progress = bpy.props.IntProperty(name="", default=0, min=0, max=generated_args['steps'], update=step_progress_update)
```
NullSenseStudio, line 89: I've found it's more performant to put similar functionality within the step callback. Strangely changing this to an empty function would still allow the UI to redraw. I assume it was redrawing more than necessary for seeing the progress bar. Allows for generating about 15%-20% faster in my experience.

operators/dream_texture.py
```diff
@@ -64,14 +83,7 @@ def execute(self, context):
...
         if generated_args['use_init_img']:
-            match generated_args['init_img_src']:
-                case 'file':
-                    init_image = scene.init_img
-                case 'open_editor':
-                    for area in screen.areas:
-                        if area.type == 'IMAGE_EDITOR':
-                            if area.spaces.active.image is not None:
-                                init_image = area.spaces.active.image
+            get_source_image(context, generated_args['init_img_src'])
```
carson-katri, line 86:
```suggestion
            init_image = get_source_image(context, generated_args['init_img_src'])
```

carson-katri (changes requested):
Nice improvements. Just one small change for img2img.

carson-katri (approved):

PR #703, opened 2023-08-19
Model Loading Improvements and Safetensors Importing

NullSenseStudio:
### Changes

Downloading models now uses diffusers own implementation as is does a great job of only getting the required files for inference. It will also handle variant files instead of a separate revision/branch for fp16 weights. SDXL will now be properly downloaded due to not blacklisting .safetensors weights.

Reimplemented loading a model in half precision and restricted it to only using cached weights. Incomplete models will now raise an error again instead of continuing to download without indication to the user.

Fixed loading of imported models and updated model importing to use diffusers own implementation and support for safetensors. @carson-katri, I'm not familiar with why the whole checkpoint conversion source was included. Were there any necessary changes involved for importing depth and inpainting models besides the configs?

### TODO
- [x] fp16 revision loading. Currently fp16 preference on by default, so I'm sure most users updating from a previous version are going to run into issues with how it works now.
- [x] Allow SDXL base and refiner to stay loaded at the same time. Reloading both pipelines from disk is not ideal if you have enough memory. I'm assuming it was implemented this way due to memory limitations on macOS and it requiring fp32, so I'll leave it as it is for MPS and CPU.
- [x] Reimplement cached pipeline invalidation for certain optimization changes.
- [x] direct checkpoint loading using `from_single_file`
- [x] Reimplement scheduler loading


carson-katri:
Regarding checkpoint conversion, it was simpler at the time to include the full script because it wasn't part of the diffusers Python package. With `download_from_original_stable_diffusion_ckpt`, that seems to be resolved though.

NullSenseStudio:
Re-added getting revision paths because the auto pipelines don't account for `local_files_only` when fetching configs. If there is an update to the model it would download only `model_index.json` and change the revision hash, causing the previously downloaded weights to be unused. I'll file a proper bug report to diffusers when I get the time.

NullSenseStudio:
Loading from checkpoint works well enough. They can be linked in preferences similar to how they are imported. Entire folders can be selected or the checkpoint files directly.
image
A few issues though. First being that models are identified by their basename in the diffusers backend, which will prevent models with the same name in different locations from being loadable. Not a huge issue but worth noting. The second issue is that killing the subprocess will break checkpoint loading until a checkpoint link has been added/removed or the addon is disable and re-enabled.

NullSenseStudio:
I've noticed ControlNet pipelines don't convert to/from a non-ControlNet pipeline with the same weights. It doesn't seem to be accounted for in the auto pipelines `from_pipe()` method. May just have to add another cache invalidation check for that.

NullSenseStudio:
SDXL base+refiner can now remain loaded together on CUDA. I've noticed around 28% less time using this with model offloading compared to reloading each pipeline when they would be used. Time savings should be even higher for cards with enough VRAM to not require model offloading. I've also extended this to keeping control nets cached between iterations, but I don't have any timings to share for that.

I've also made it possible to save imported weights in half precision, and importing will now show its progress instead of freezing Blender until it's done.
blender_J7oiG3xlpe

generator_process/models/model_config.py
```diff
@@ -0,0 +1,60 @@
...
+    AUTO_DETECT = "auto-detect"
+    STABLE_DIFFUSION_1 = "v1"
+    STABLE_DIFFUSION_2_BASE = "v2 (512, epsilon)"
+    STABLE_DIFFUSION_2 = "v2 (768, v_prediction)"
+    STABLE_DIFFUSION_2_DEPTH = "v2 (depth)"
+    STABLE_DIFFUSION_2_INPAINTING = "v2 (inpainting)"
+    STABLE_DIFFUSION_XL_BASE = "XL (base)"
+    STABLE_DIFFUSION_XL_REFINER = "XL (refiner)"
+    CONTROL_NET_1_5 = "1.5 (ControlNet)"
+    CONTROL_NET_2_1 = "2.1 (ControlNet)"
```
carson-katri, line 16: Do we need a separate case here for XL ControlNet models?

NullSenseStudio: Any XL ControlNet models I can find are only in diffusers format. I'm not aware of any yaml config files that can be used for converting or that I'm familiar enough with the parameters to mock one up.

generator_process/models/checkpoint.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+from .model_config import ModelConfig
+
+
+@dataclass
+class Checkpoint:
+    path: str
+    config: ModelConfig | str | None
```
carson-katri, line 9: I had to add a `__hash__` function for this to work:
```suggestion
    config: ModelConfig | str | None
    
    def __hash__(self):
        return hash((self.path, self.config))
```
Otherwise, an `unhashable type` error occurred [here in the `cache_check` decorator](https://github.com/carson-katri/dream-textures/pull/703/files#diff-cbf852790591d9974d18a3d84f89f66fc64e3803ac294f17f5c1786838ece249R40)

NullSenseStudio: I opted for making the class frozen to allow it to generate the hash function. Seems best this way since it is used as a key.

carson-katri (approved):

PR #707, opened 2023-09-18
DirectML Out of Memory Retry

NullSenseStudio:
Allows DML execution in submodules to retry running when it runs out of memory. Sometimes this can allow generation to finish, but often it'll occur again on a later step and not recover from it.

Found and fixed some DML issues caused from merging the last branch.

It can run SDXL, but may require offloading depending on what GPU is used. Using half VAE tiling is also effective and maintains good decoded image accuracy.

### DIFF
--- a/generator_process/actions/choose_device.py
+++ b/generator_process/actions/choose_device.py
@@ -1,3 +1,4 @@
+import importlib.util
 import sys
 
 def choose_device(self, optimizations) -> str:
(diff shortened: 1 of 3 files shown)


NullSenseStudio:
Found it helps quite a bit to clear the traceback before retrying. Now it'll more often recover and will usually finish 25 steps of SDXL without requiring offloading. The decoding stage still doesn't complete on my GPU without offloading or tiling though.

That should be it for this PR.

carson-katri (approved):

PR #711, opened 2023-10-09
Release v0.3.0

carson-katri:

PR #716, opened 2023-10-22
Check for refiner before unpacking tuple from `load_model`

carson-katri:
`load_model` only returns a tuple when `sdxl_refiner_model` is not `None`.

Issue described in more detail here: https://github.com/carson-katri/dream-textures/pull/711#issuecomment-1774135168


NullSenseStudio (approved):
LGTM

PR #725, opened 2023-11-05
Support seamless for latest diffusers

carson-katri:
Some [changes to `LoRACompatibleConv`](https://github.com/huggingface/diffusers/blame/2b23ec82e898aa1f0e172da6ef054631634b643d/src/diffusers/models/lora.py#L232) in diffusers broke seamless generation. This patches `LoRACompatibleConv.forward` to avoid using the functional `conv2d` (which will ignore the padding mode) and instead use our custom `_conv_forward_asymmetric`.


NullSenseStudio (approved):
LGTM

It won't properly support LoRA with seamless axes as it is, but it can be revised if LoRA integration is ever added in the future.

PR #730, opened 2023-11-06
Fix Generating from Opened Blend File

### LINKED ISSUE #729: Generate is Disabled When Opening a Blend File
### Description

Generating any image is disabled when opening a .blend file, starting Blender directly and then using its `File > Open...` alternatively does not. Releasing the generator doesn't help.

The console gets spammed with `'str' object has no attribute 'model_type'`. I've tracked this down to https://github.com/carson-katri/dream-textures/blob/4f210c51edd490f51ac31653a5fc3127fefb1e56/ui/panels/dream_texture.py#L368-L369
This actual error is occurring here https://github.com/carson-katri/dream-textures/blob/4f210c51edd490f51ac31653a5fc3127fefb1e56/diffusers_backend.py#L263-L268
For some reason `model_lookup` is not being updated and just returns the model id string.

It can be made to update by installing another model: either downloading, importing, or linking.

### Steps to Reproduce

1. open Blender and save a new .blend file
2. close Blender
3. open .blend file from file explorer (NOT from within Blender `File > Open...` / `File > Open Recent`)

### Expected Behavior

Generate button to not be disabled.

### Addon Version

Windows (CUDA)

### GPU

NVIDIA

NullSenseStudio:
fixes #729

### DIFF
--- a/generator_process/future.py
+++ b/generator_process/future.py
@@ -64,7 +64,7 @@ class Future:
     def _run_on_main_thread(self, func):
         try:
             import bpy
-            bpy.app.timers.register(func)
+            bpy.app.timers.register(func, persistent=True)
         except:
             func()
 
--- a/preferences.py
+++ b/preferences.py
@@ -123,7 +123,7 @@ class model_lookup:
 
     @classmethod
     def get(cls, item):
-        return cls._models.get(item, item)
+        return cls._models.get(item, None)
 
 def fetch_installed_models(blocking=True):
     def on_done(future):
--- a/ui/panels/dream_texture.py
+++ b/ui/panels/dream_texture.py
@@ -365,6 +365,4 @@ def actions_panel(sub_panel, space_type, get_prompt):
                 for i, line in enumerate(e.args[0].split('\n')):
                     error_box.label(text=line, icon="ERROR" if i == 0 else "NONE")
                 e._draw(prompt, context, error_box)
-            except Exception as e:
-                print(e)
     return ActionsPanel


generator_process/future.py
```diff
@@ -64,7 +64,7 @@ def cancel(self):
     def _run_on_main_thread(self, func):
         try:
             import bpy
-            bpy.app.timers.register(func)
+            bpy.app.timers.register(func, persistent=True)
```
NullSenseStudio, line 67: Makes sure that model fetching completes when loading a file. Often when loading a file first it would fetch the models, then it gets queued in this timer, then the file actually loads and discards the timer. On rare occasions model fetching would complete along with the timer before the .blend file loads.

preferences.py
```diff
@@ -123,7 +123,7 @@ class model_lookup:
 
     @classmethod
     def get(cls, item):
-        return cls._models.get(item, item)
+        return cls._models.get(item, None)
```
NullSenseStudio, line 126: We don't want to default to returning the model id string in case if the id is invalid or the models haven't loaded yet.

ui/panels/dream_texture.py
```diff
@@ -365,6 +365,4 @@ def draw(self, context):
                 for i, line in enumerate(e.args[0].split('\n')):
                     error_box.label(text=line, icon="ERROR" if i == 0 else "NONE")
                 e._draw(prompt, context, error_box)
-            except Exception as e:
-                print(e)
```
NullSenseStudio, line 369: These exceptions can be safely propagated since its at the end of the draw function, now it can properly show traceback info on errors it encounters.

carson-katri (approved):
Good catch

PR #737, opened 2023-11-09
Add default nodes when creating a new node tree using new_engine_node_tree

GottfriedHofmann:
When creating new node trees, usually a functional minimal sample of a node tree is already created (Eevee, Cycles, Geometry Nodes, Compositor Nodes). This is not possible directly when the node tree type is defined via Python but Dream Textures has a secondardy operator new_engine_node_tree that is exposed in the render engine settings. Using this operator, we can simulate the default Blender behavior.

For the default nodes, I have selected Stable Diffusion and Output because they are the bare minimum render properties because that way the user will not run into the error of resolution mismatch between render engine and the stable diffusion output.

grafik


engine/engine.py
```diff
@@ -87,7 +87,33 @@ class NewEngineNodeTree(bpy.types.Operator):
...
+            if nt.bl_idname == 'DreamTexturesNodeTree' and 'Dream Textures Node Editor' in nt.name:
+                node_trees.append(nt)
+        if node_trees != []:
+            # assume that the newly created node tree is the last one in the list
+            node_tree = node_trees[-1]
+            node_sd = node_tree.nodes.new(type="dream_textures.node_stable_diffusion")
+            node_sd.location = (200, 200)
+            node_out = node_tree.nodes.new(type="NodeGroupOutput")
+            node_out.location = (400, 200)
+            node_tree.links.new(node_sd.outputs['Image'], node_out.inputs[''])
```
carson-katri, line 107: I think this connects the image to the virtual socket on the Group Output. You might need to add an image socket to it first, then link to that socket.

GottfriedHofmann: It worked without but of course it should reflect what the user would get if they did it themselves so adding a color socket first is the cleaner solution for sure 👍🏻

engine/engine.py
```diff
@@ -87,7 +87,34 @@ class NewEngineNodeTree(bpy.types.Operator):
...
-        bpy.ops.node.new_node_tree(type="DreamTexturesNodeTree")
+        # TODO: Get the name of the default node tree from a config file
+        # TODO: type is deprecated https://docs.blender.org/api/current/bpy.types.NodeTree.html 
+        #       When testing the type of the resulting node tree using  bpy.data.node_groups['Dream Textures Node Tree'].type it is '' because of that
+        bpy.ops.node.new_node_tree(type="DreamTexturesNodeTree", name="Dream Textures Node Editor")
+        # Get the newly generated node tree
+        node_trees = []
+        for nt in bpy.data.node_groups:
+            # the correct identifier is bl_idname nowadays
+            if nt.bl_idname == 'DreamTexturesNodeTree' and 'Dream Textures Node Editor' in nt.name:
```
carson-katri, line 98: Immediately after constructing the node tree, it should be the last group in `bpy.data.node_groups`. Could it be accessed with `bpy.data.node_groups[-1]` instead of searching by name?

This is done on L102, but only after filtering by name/type.

carson-katri (changes requested):
One small suggestion to simplify the new tree lookup.

engine/engine.py
GottfriedHofmann: It works and simplifies the code quiet a bit, thanks 👍🏻

engine/engine.py
```diff
@@ -87,7 +87,27 @@ class NewEngineNodeTree(bpy.types.Operator):
...
+        # TODO: Get the name of the default node tree from a config file
+        # TODO: type is deprecated https://docs.blender.org/api/current/bpy.types.NodeTree.html 
+        #       When testing the type of the resulting node tree using  bpy.data.node_groups['Dream Textures Node Tree'].type it is '' because of that
+        bpy.ops.node.new_node_tree(type="DreamTexturesNodeTree", name="Dream Textures Node Editor")
+        # Get the newly generated node tree
+        node_tree = bpy.data.node_groups[-1]
+        node_sd = node_tree.nodes.new(type="dream_textures.node_stable_diffusion")
+        node_sd.location = (200, 200)
+        node_out = node_tree.nodes.new(type="NodeGroupOutput")
+        node_tree.outputs.new('NodeSocketColor','Image')
```
carson-katri, line 99: Missed this before, `outputs` and `inputs` were removed in Blender 4. Could you add a check here for the `bpy.app.version`, and in Blender 4.0 use the [`.interface.new_socket(...)`](https://docs.blender.org/api/current/bpy.types.NodeTreeInterface.html) API?

engine/engine.py
```diff
@@ -87,7 +87,27 @@ class NewEngineNodeTree(bpy.types.Operator):
...
 
     def execute(self, context):
-        bpy.ops.node.new_node_tree(type="DreamTexturesNodeTree")
+        # TODO: Get the name of the default node tree from a config file
+        # TODO: type is deprecated https://docs.blender.org/api/current/bpy.types.NodeTree.html 
+        #       When testing the type of the resulting node tree using  bpy.data.node_groups['Dream Textures Node Tree'].type it is '' because of that
+        bpy.ops.node.new_node_tree(type="DreamTexturesNodeTree", name="Dream Textures Node Editor")
+        # Get the newly generated node tree
+        node_tree = bpy.data.node_groups[-1]
+        node_sd = node_tree.nodes.new(type="dream_textures.node_stable_diffusion")
```
carson-katri, line 96: It might be worthwhile to pass `NodeStableDiffusion.bl_idname` as the type in case the ID changes in the future.

engine/engine.py
```diff
@@ -87,7 +87,27 @@ class NewEngineNodeTree(bpy.types.Operator):
...
+        node_sd.location = (200, 200)
+        node_out = node_tree.nodes.new(type="NodeGroupOutput")
+        node_tree.outputs.new('NodeSocketColor','Image')
+        node_out.location = (400, 200)
+        node_tree.links.new(node_sd.outputs['Image'], node_out.inputs['Image'])
+        node_props = node_tree.nodes.new(type="dream_textures.node_render_properties")
```
carson-katri, line 102: This could use `NodeRenderProperties.bl_idname`.

engine/engine.py
GottfriedHofmann: Added this to latest iteration but was not able to test yet.

carson-katri (approved):

PR #745, opened 2023-11-16
Blender 4.0 and Command Line Rendering Fixes

NullSenseStudio:
Prevents deadlocks by checking when sections are running on the main thread to continue executing on the same thread rather than registering a timer and waiting on that. Forces requests to the subprocess to run in blocking mode while command line rendering is occurring, and since forcing blocking mode makes the returned Future complete before callbacks get added the done/exception callbacks will run immediately in that case.


engine/engine.py
```diff
@@ -18,6 +19,7 @@ class DreamTexturesRenderEngine(bpy.types.RenderEngine):
     bl_label = "Dream Textures"
     bl_use_preview = False
     bl_use_postprocess = True
+    bl_use_gpu_context = actor.main_thread_rendering
```
NullSenseStudio, line 22: For some reason `bl_use_gpu_context = True` allows for rendering annotations in background mode, but freezes under normal use. And `bl_use_gpu_context = False` works fine under normal use, but freezes while in background mode. So `bl_use_gpu_context = actor.main_thread_rendering` is the best that I'll get this to work.

NullSenseStudio:
Forgot to mention that I've made it cut down on some spam caused by the backend default not existing when you don't have any extra backend installed, and fixed a couple of issues for running on Blender 4.0 too.

carson-katri (approved):

PR #788, opened 2024-04-21
Zip Dependencies with Long Paths

NullSenseStudio:
Currently Windows users struggle to install packaged releases due to errors caused by long paths, even when long paths are enabled for their system. Once installed a certain file has an entire path length of 259 characters when not including the Windows username. This became more of a noticeable issue in the last release due to a path in the transformers module growing by 11 characters.

To help prevent this issue, this PR zips the transformers module separately and enables it to be loaded like normal. Now the max path length is 225, so installing shouldn't be an issue as long as the username isn't more than 34 characters or another module grows in path length. This method is a little limiting as it only supports zipping modules that use .py files, no .dll or other special files. The new longest path belongs to torch, which is not compatible to shorten the path length further.

### DIFF
--- a/.github/workflows/package-release.yml
+++ b/.github/workflows/package-release.yml
@@ -45,6 +45,9 @@ jobs:
         shell: bash
         run: 'python -m pip install -r requirements/win-linux-cuda.txt --no-cache-dir --target .python_dependencies'
         working-directory: dream_textures
+      - name: Zip dependencies with long paths
+        shell: bash
+        run: 'python ./dream_textures/scripts/zip_dependencies.py'
       - name: Archive Release
         uses: thedoctor0/zip-release@main
         with:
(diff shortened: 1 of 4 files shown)


NullSenseStudio:
The described issue encountered while installing:

`FileNotFoundError: [Errno 2] No such file or directory: '...\\dream_textures\\.python_dependencies\\transformers\\models\\deprecated\\trajectory_transformer\\__pycache__\\convert_trajectory_transformer_original_pytorch_checkpoint_to_pytorch.cpython-310.pyc'`

This error message is only noticeable with the system console window open. In my experience it just minimizes the preferences window so there is no proper feedback that this error occurred and the user continues on thinking it's just complete. Next when trying to enable the addon a new error will appear in the preferences UI:

`ModuleNotFoundError: No module named 'dream_textures.api'`

And in older versions it could appear as:

``ModuleNotFoundError: No module named 'dream_textures.prompt_engineering'``

carson-katri (approved):
This is great, thanks!

PR #794, opened 2024-05-17
Only check for issues with 'bug' label

carson-katri:
This option should only run the stale bot on issues with the `bug` label. If any other labels are present, it will not mark the issue as stale.

The `confirmed` label, or a specifier label such as `AMD`, `NVIDIA`, `render pass`, can be used to disable the stale bot for an issue.


NullSenseStudio (approved):
LGTM

PR #796, opened 2024-05-19
shm.dll fix

NullSenseStudio:
The recent PyTorch 2.3.0 version has a new MKL dependency on Windows that isn't being loaded due to the uncommon method of installing to a target directory.

Fixes `OSError: [WinError 126] The specified module could not be found. Error loading "...\dream_textures\.python_dependencies\torch\lib\shm.dll" or one of its dependencies.`


carson-katri (approved):
Thanks for tracking this down

PR #797, opened 2024-05-22
Small fix for default node setup

GottfriedHofmann:
Bugfix: When creating the default node setup, use socket_type as keyword argument.


carson-katri (approved):
Thanks

PR #801, opened 2024-06-03
Fix render pass image util to flatten array

carson-katri:
This was failing before, as `foreach_set` is expecting a flat array instead of a 2D array of `(width * height, channels)`

### DIFF
--- a/image_utils.py
+++ b/image_utils.py
@@ -689,7 +689,10 @@ def np_to_render_pass(
         array = to_dtype(array, dtype)
     if top_to_bottom:
         array = np.flipud(array)
-    render_pass.rect.foreach_set(array.reshape(-1))
+    if BLENDER_VERSION >= (4, 1, 0):
+        render_pass.rect.foreach_set(array.reshape(-1))
+    else:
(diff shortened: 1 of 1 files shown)


image_utils.py
```diff
@@ -689,7 +689,7 @@ def np_to_render_pass(
         array = to_dtype(array, dtype)
     if top_to_bottom:
         array = np.flipud(array)
-    render_pass.rect.foreach_set(array.reshape(-1, render_pass.channels))
+    render_pass.rect.foreach_set(array.reshape(-1))
```
NullSenseStudio, line 692:
```suggestion
    if BLENDER_VERSION >= (4, 1, 0):
        render_pass.rect.foreach_set(array.reshape(-1))
    else:
        render_pass.rect.foreach_set(array.reshape(-1, render_pass.channels))
```
Looks like 3.6 and 4.0 (untested for anything earlier but I expect the same) require 2D, and 4.1 requires 1D.

NullSenseStudio (changes requested):

image_utils.py
carson-katri: Ah, that explains it. Thanks for checking on that.

PR #803, opened 2024-06-06
Update linux-rocm.txt

B4rr3l-Rid3r:
Updated requirements for ROCm and Torch

### DIFF
--- a/requirements/linux-rocm.txt
+++ b/requirements/linux-rocm.txt
@@ -5,8 +5,8 @@ accelerate
 huggingface_hub
 controlnet-aux==0.0.7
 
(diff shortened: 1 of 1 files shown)


carson-katri:
Thanks!

aaaaaaaaargh:
Quick question about this PR. How is this supposed to work? As far as I can see there is no stable version of Torch that fulfills this requirement, so 2.4.0 stable non-rocm version is being downloaded instead.

In order to get it right I hat do change `linux-rocm.txt` to this:

```
--extra-index-url https://download.pytorch.org/whl/nightly/rocm6.1/
torch==2.5.0.dev20240811+rocm6.1
```

Obviously not a very elegant solution, but maybe I'm missing something here.

PR #808, opened 2024-08-11
Fix Blender 4.1 Render Pass

NullSenseStudio:
Made a similar change in `image_utils.render_pass_to_np()` that was previously added to `image_utils.np_to_render_pass()` due to a change in Blender 4.1 no longer expecting a channel dimension when reading/writing to a render pass.


carson-katri (approved):
Thanks!

PR #810, opened 2024-08-12
Remove `ModelFilter` for `huggingface_hub` deprecations

carson-katri:
This fixes the search in response to `huggingface_hub` deprecating `ModelFilter`.

Closes #787

PR #811, opened 2024-08-12
Remove `nightly` prefix from ROCm `extra-index-url`

### LINKED ISSUE #809: linux-rocm requirements don't pull torch-rocm
### Description

As already mentioned in #803, manually installing the dependencies using the provided linux-rocm requirements file is not working as expected.

### Steps to Reproduce

1. Clone the repository in your scripts/addons folder as described in the documentation.
2. Run `/usr/share/blender/4.1/python/bin/python3.11 -m pip install -r requirements/linux-rocm.txt --target .python_dependencies` in the checked out folder

### Expected Behavior

A rocm version of PyTorch (>=2.3.1) should have been downloaded but instead it is downloading the default one withtout rocm support.

This is due to `requirements/linux-rocm.txt` referring to the extra-index-url `https://download.pytorch.org/whl/nightly/rocm6.1/` which is only containing nightly versions and from my understanding the version constraint `>=2.3.1` does not apply. Changing the extra-index-url to `https://download.pytorch.org/whl/rocm6.1/` seems to fix the problem and I couldn't find any issues when using the addon in blender.

(Also, as a side note, the dropdown in the bug reporting form should be updated because the current gen AMD GPUs are supported now)

### Addon Version

Other (Built from source)

### GPU

AMD (unsupported for local generation)

carson-katri:
Closes #809

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -47,11 +47,19 @@ body:
       multiple: false
       options:
         - Windows (CUDA)
-        - Windows (DreamStudio-only)
-        - macOS (Intel)
+        - Windows (DirectML)
         - macOS (Apple Silicon)
-        - macOS (Apple Silicon, DreamStudio-only)
         - Other (Built from source)
+    validations:
+      required: true
+  - type: dropdown
+    id: blender-version
+    attributes:
+      label: Blender Version
+      multiple: false
+      options:
+        - Blender 4.1+
+        - Blender 3.6 - 4.0
     validations:
       required: true
   - type: dropdown
@@ -63,7 +71,7 @@ body:
       options:
         - NVIDIA
         - NVIDIA 16 Series
-        - AMD (unsupported for local generation)
+        - AMD
         - Apple Silicon
         - Other
     validations:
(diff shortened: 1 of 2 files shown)


requirements/linux-rocm.txt
```diff
@@ -5,7 +5,7 @@ accelerate
 huggingface_hub
```
aaaaaaaaargh, line 5: I had to downgrade as well in order to make huggingface model downloads working:

`huggingface_hub==0.22.0`

But I'm not entirely sure if this PR is the right place to do that because I'm pretty sure other platforms are affected as well.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -63,7 +71,7 @@ body:
       options:
         - NVIDIA
         - NVIDIA 16 Series
-        - AMD (unsupported for local generation)
+        - AMD
```
aaaaaaaaargh, line 74: thank you! :+1:

requirements/linux-rocm.txt
carson-katri: I fixed that properly in #810

aaaaaaaaargh (approved):
just double checked with my diff and that's pretty much what I got :+1:

requirements/linux-rocm.txt
aaaaaaaaargh: I see, you migrated to the API. That's much better.

PR #812, opened 2024-08-13
Use id rather than modelId from Hugging Face

osanseviero:
For context, modelId is getting deprecated https://github.com/huggingface/huggingface_hub/issues/2408


carson-katri:
I believe this was already fixed in #813

